PR #20, opened 2024-08-06
Feat(dataProvider): Add supabase to fakerest filter adapter

jonathan-marmelab:
## Problem

We want to support both supabase and fake rest as auth and data providers. However, supabase list filters are not compatible with fake rest's ones.

Filters to support:
- [x] eq
- [x] neq
- [X] in
- [x] cs
- [x] is
- [x] not.is
- [x] lt
- [x] lte
- [x] gt
- [x] gte
- [x] or
- [x] ilike
- [x] like

## Solution

Add an adapter from supabase filters to fake rest filters.

## How To Test

Replace `dataProvider` and `authProvider` by fake rest one's in `root/CRM.tsx`

## Additional Checks

### DIFF
--- a/src/providers/fakerest/internal/transformFilter.ts
+++ b/src/providers/fakerest/internal/transformFilter.ts
@@ -1,5 +1,6 @@
 import { transformContainsFilter } from './transformContainsFilter';
 import { transformInFilter } from './transformInFilter';
+import { transformOrFilter } from './transformOrFilter';
 
 export function transformFilter(filter: Record<string, any>) {
     const transformedFilters: Record<string, any> = {};
@@ -28,6 +29,12 @@ export function transformFilter(filter: Record<string, any>) {
         if (key.endsWith('@cs')) {
             transformedFilters[`${key.slice(0, -3)}`] =
                 transformContainsFilter(value);
+            continue;
+        }
+
+        // Search query
+        if (key.endsWith('@or')) {
+            transformedFilters['q'] = transformOrFilter(value);
             continue;
         }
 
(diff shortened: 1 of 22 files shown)


src/providers/fakerest/internal/transformFilter.ts
```diff
@@ -0,0 +1,54 @@
...
+            transformedFilters[
+                `${key.substring(0, lastIndexOfAt)}_${key.substring(lastIndexOfAt + 1)}`
+            ] = value;
+            continue;
+        }
+
+        if (key.endsWith('@is')) {
+            transformedFilters[`${key.slice(0, -3)}_eq`] = value;
+            continue;
+        }
```
slax57, line 26: **question**: have you checked how this fares with nullish values? I recall having issues with fakerest not being able to filter null values correctly...

src/providers/fakerest/internal/transformInFilter.ts
```diff
@@ -0,0 +1,20 @@
+const IN_FILTER_REGEX = /^\([a-z0-9-]+(,[a-z0-9-]+)*\)$/i;
```
slax57, line 1: Same comment as for contains

slax57 (changes requested):
Otherwise looks very nice, bravo :clap:

src/providers/fakerest/internal/transformFilter.ts
jonathan-marmelab: It seemed to work during my tests

slax57 (approved):

PR #23, opened 2024-08-06
Fix(setting): Update current user information in supabase

arimet:
## Problem

Currently, when a user modifies their profile, the information is reset at logout. 

## Solution

Modify the associated user in supabase

### DIFF
--- a/src/settings/SettingsPage.tsx
+++ b/src/settings/SettingsPage.tsx
@@ -46,9 +46,7 @@ export const SettingsPage = () => {
             }
             return dataProvider.salesUpdate(identity.id, data);
         },
-        onSuccess: data => {
-            // Update local user
-            localStorage.setItem(USER_STORAGE_KEY, JSON.stringify(data));
+        onSuccess: () => {
             refetchIdentity();
             refetchUser();
             setEditMode(false);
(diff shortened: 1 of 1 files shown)


src/settings/SettingsPage.tsx
```diff
@@ -9,62 +9,67 @@ import {
...
+        mutationKey: ['signup'],
+        mutationFn: async (data: SalesFormData) => {
+            if (!identity) {
+                throw new Error('Record not found');
+            }
+            return dataProvider.salesUpdate(identity.id, data);
+        },
+        onSuccess: data => {
+            // Update local user
+            localStorage.setItem(USER_STORAGE_KEY, JSON.stringify(data));
```
jonathan-marmelab, line 51: **question:** Is local storage still required ? As we do not use it we supabase ?

jonathan-marmelab (changes requested):

src/settings/SettingsPage.tsx
```diff
@@ -9,62 +9,65 @@ import {
...
+
+    const { mutate } = useMutation({
+        mutationKey: ['signup'],
+        mutationFn: async (data: SalesFormData) => {
+            if (!identity) {
+                throw new Error('Record not found');
+            }
+            return dataProvider.salesUpdate(identity.id, data);
+        },
+        onSuccess: data => {
```
slax57, line 49: linter warning at this line. Otherwise, seems to look good.

PR #25, opened 2024-08-06
Feat(task): Associate task to an sales_id + improve documentation

arimet:
## Problem

- Actually, you can only see Task that are associated with a contact that belongs to me.
- Currently documentation is out of date 

## Solution

- Associate Task to a sales_id. Display all Task for this sales in the Dashboard.
- Update documentation  


README.md
```diff
@@ -99,4 +98,143 @@ have to create the following secrets:
...
+
+const App = () => (
+    <CRM 
+        title="Custom CRM Title" 
+        logo="custom-logo.png" 
+    />
+);
+
+export default App;
+```
```
slax57, line 121: The improvements to the docs look good, but did you intend to include them in this PR? Neither the title nor the description speaks about that...

src/dashboard/TasksListFilter.tsx
```diff
@@ -29,10 +29,10 @@ export const TasksListFilter = ({
             sort: { field: 'due_date', order: 'ASC' },
             filter: {
                 ...filter,
-                'contact_id@in': `(${contacts.map(contact => contact.id)})`,
+                'sales_id@in': `(${identity?.id})`,
```
slax57, line 32: why keep the `in` operator here? :thinking: 

```suggestion
                sales_id: identity?.id,
```

slax57 (commented):
Also, needs rebase apparently

jonathan-marmelab (commented):
**nit-pick:** Since the readme file is getting large (and we sill have documentationto have), maybe we should split it into mutliple doc files under a doc directory and use the readme as table of contents ?

supabase/migrations/20240806124555_task_sales_id.sql
```diff
@@ -0,0 +1 @@
+alter table "public"."tasks" add column "sales_id" bigint;
```
jonathan-marmelab, line 1: **issue:** You forgot to update the `Task` type in `types.ts` to reflect the change

jonathan-marmelab (changes requested):

arimet:
> **nit-pick:** Since the readme file is getting large (and we sill have documentationto have), maybe we should split it into mutliple doc files under a doc directory and use the readme as table of contents ?

Okay, but there's still a lot to document, and there's a ticket for that. I'm in favor of doing it then.

src/types.ts
```diff
@@ -126,6 +126,7 @@ export interface Task extends RaRecord {
     text: string;
     due_date: string;
     done_date?: string | null;
+    sales_id: Identifier;
```
slax57, line 129: Should be `sales_id?: Identifier;` since the column does not have the `not null` constraint in DB?

jonathan-marmelab (approved):

PR #26, opened 2024-08-06
Feat(auth): Add reset-password for Sales

arimet:
## Problem

As a user, I can't reset my password If I forgot it

## Solution

Add reset password and set password page. Set a recovery email and add documentation about it. 

image
image
 
## How To Test Locally

You will need to restart your local supabase instance to apply the latest config.

```sh
make stop
make start
```

You can then use InBucket (available at http://localhost:54324/) to see the password-reset emails and click on the _Reset my password_ button.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,3 +100,31 @@ SUPABASE_ACCESS_TOKEN: Your personal access token, can be found at https://supab
 SUPABASE_DB_PASSWORD: Your supabase database password
 SUPABASE_PROJECT_ID: Your supabase project id
 ```
+
+## Supabase
+
+### Password Reset When Forgotten
+If users forgot their password, they can request for a reset. Atomic CRM handle it for you. All the magic is done by the custom route `/forgot-password` and `/reset-password` that you can find in the `CRM` component.
+
+If you want to update default configuration: 
+
+1. Go to your `config.toml` file
+2. In `[auth]` section set `site_url` to  your application URL
+3. In `[auth]`, add the following URL in the `additional_redirect_urls = [{APPLICATION_URL}}/auth-callback"]`
+4. Add an `[auth.email.templace.recovery]` section with the following option
+```
+[auth.email.template.recovery]
+subject = "Reset Password"
+content_path = "./supabase/templates/recovery.html"
+```
+
+In Recovery email template set the `auth-callback` redirection
+
+```HTML
+<html>
+  <body>
+    <h2>Reset Password</h2>
+    <p><a href="{{ .ConfirmationURL }}/auth-callback">Reset your password</a></p>
+  </body>
+</html>
+```
--- a/README.md
+++ b/README.md
@@ -110,14 +110,16 @@ GIT_USER_EMAIL: The deploy user's email
 ## Supabase
 
 ### Password Reset When Forgotten
-If users forgot their password, they can request for a reset. Atomic CRM handle it for you. All the magic is done by the custom route `/forgot-password` and `/reset-password` that you can find in the `CRM` component.
+If users forgot their password, they can request for a reset. Atomic CRM handle it for you. All the magic is done by the custom route `/forgot-password` and `/set-password` that you can find in the `CRM` component.
 
 If you want to update default configuration: 
+
+### Local development
 
 1. Go to your `config.toml` file
 2. In `[auth]` section set `site_url` to  your application URL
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -105,4 +105,31 @@ You will also need to configure the following environment variables to deploy to
...
 GIT_USER_NAME: The deploy account's name
 GIT_USER_EMAIL: The deploy user's email
-```
\ No newline at end of file
+```
+
+## Supabase
+
+### Password Reset When Forgotten
+If users forgot their password, they can request for a reset. Atomic CRM handle it for you. All the magic is done by the custom route `/forgot-password` and `/reset-password` that you can find in the `CRM` component.
```
slax57, line 113: nitpick (non-blocking):

```suggestion
If users forgot their password, they can request for a reset. Atomic CRM handle it for you. All the magic is done by the custom route `/forgot-password` and `/set-password` that you can find in the `CRM` component.
```

README.md
```diff
@@ -105,4 +105,31 @@ You will also need to configure the following environment variables to deploy to
...
+
+If you want to update default configuration: 
+
+1. Go to your `config.toml` file
+2. In `[auth]` section set `site_url` to  your application URL
+3. In `[auth]`, add the following URL in the `additional_redirect_urls = [{APPLICATION_URL}}/auth-callback"]`
```
slax57, line 119: This will work for local environment, but how did you manage the prod environment?

README.md
```diff
@@ -105,4 +105,31 @@ You will also need to configure the following environment variables to deploy to
...
+4. Add an `[auth.email.templace.recovery]` section with the following option
```
slax57, line 120:
```suggestion
4. Add an `[auth.email.template.recovery]` section with the following option
```

supabase/config.toml
```diff
@@ -69,9 +69,9 @@ file_size_limit = "50MiB"
 enabled = true
 # The base URL of your website. Used as an allow-list for redirects and for constructing URLs used
 # in emails.
-site_url = "http://127.0.0.1:3000"
+site_url = "http://127.0.0.1:5173"
 # A list of *exact* URLs that auth providers are permitted to redirect to post authentication.
-additional_redirect_urls = ["https://127.0.0.1:3000"]
+additional_redirect_urls = ["https://127.0.0.1:5173/auth-callback"]
```
slax57, line 74: I had issues after chaining 2 password resets, because the 2nd time the URL is `https://127.0.0.1:5173/` instead of `https://localhost:5173/`, and for some reason then Supabase redirects to `https://localhost:5173//auth-callback` instead of `https://localhost:5173/auth-callback` (notice the extra `/`). This leads to a Not Found screen.

Maybe this can all be fixed by changing the config to:

```toml
site_url = "http://localhost:5173"
# A list of *exact* URLs that auth providers are permitted to redirect to post authentication.
additional_redirect_urls = ["https://localhost:5173/auth-callback"]
```

supabase/templates/recovery.html
```diff
@@ -0,0 +1,85 @@
...
+            <div class="content">
+                <p>Hello,</p>
+                <p>To reset your password, follow the link below:</p>
+                <a href="{{ .ConfirmationURL }}/auth-callback" class="button"
+                    >Reset my password</a
+                >
+                <p>
+                    <strong>Warning:</strong> If the password reset request did
+                    not originate from you, please contact us immediately as
+                    this may be a fraudulent attempt.
```
slax57, line 69: thought (non-blocking): it could be nice to be more explicit about how the users are supposed to contact us. Dedicated email? GH Issue? Or maybe remove this whole sentence as it is not really applicable at the moment.

supabase/templates/recovery.html
```diff
@@ -0,0 +1,85 @@
...
+                </p>
+                <p>Keep up the great work!</p>
+                <p>The Atomic CRM team</p>
+            </div>
+            <div class="footer">
+                <p>
+                    If you need help, feel free to
+                    <a
+                        href="mailto:support@react-admin-enterprise-edition.zendesk.com"
```
slax57, line 78: This link is dedicated to the paid React Admin Enterprise support. I don't think we should include it there as it is misleading.

slax57 (changes requested):

supabase/templates/recovery.html
arimet: The wording of the mail has been given and approved it seems to me

slax57 (approved):

PR #27, opened 2024-08-06
Add fake rest provider

jonathan-marmelab:
## Problem

We want to support both supabase and fake rest as data providers

## Solution

Add support for fake rest data provider that matches supabase data provider methods
 
## How To Test

Use fake rest data providers and auth providers in CRM component

## Additional Checks


src/providers/fakerest/dataProvider.ts
```diff
@@ -20,10 +20,11 @@ import { getActivityLog } from '../commons/activity';
 import { getCompanyAvatar } from '../commons/getCompanyAvatar';
 import { getContactAvatar } from '../commons/getContactAvatar';
 import { CrmDataProvider } from '../types';
+import { authProvider, USER_STORAGE_KEY } from './authProvider';
 import generateData from './dataGenerator';
 import { withSupabaseFilterAdapter } from './internal/supabaseAdapter';
 
-const baseDataProvider = fakeRestDataProvider(generateData(), true, 300);
+const baseDataProvider = fakeRestDataProvider(generateData(), false, 300);
```
slax57, line 27: question: why that change?

src/providers/fakerest/dataProvider.ts
```diff
@@ -241,6 +242,16 @@ export const dataProvider = withLifecycleCallbacks(
...
+            afterSave: async data => {
+                const currentUser = await authProvider.getIdentity?.();
+                if (currentUser?.id === data.id) {
+                    localStorage.setItem(
+                        USER_STORAGE_KEY,
+                        JSON.stringify(data)
+                    );
+                }
+                return data;
+            },
```
slax57, line 254: suggestion: add a little comment to explain why this is necessary? (it certainly is not obvious for me at least)

src/providers/fakerest/dataProvider.ts
jonathan-marmelab: Too much logs in the console, debug is awful with this feature to true

slax57 (approved):
Couldn't find anything wrong with the code or by testing, nice work!!

PR #42, opened 2024-08-08
Fix(ux): Reduce initial loading time

jonathan-marmelab:
## Problem

Initial load time is too long if app is not initialized

## Solution

Avoid `ra-supabase` permission check if app is not initialized

## How To Test

Reset you DB and create a new account

## Additional Checks

### DIFF
--- a/src/providers/supabase/authProvider.ts
+++ b/src/providers/supabase/authProvider.ts
@@ -36,9 +36,20 @@ const baseAuthProvider = supabaseAuthProvider(supabase, {
 });
 
 export async function getIsInitialized() {
-    const { data } = await supabase.from('init_state').select('is_initialized');
+    if (getIsInitialized._is_initialized_cache == null) {
+        const { data } = await supabase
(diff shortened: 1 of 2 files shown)


src/dataProvider.ts
```diff
@@ -121,6 +121,9 @@ const dataProviderWithCustomMethods = {
             throw new Error('Failed to create account');
         }
 
+        // Update the is initialized cache
+        getIsInitialized._is_initialized_cache = true;
```
fzaninotto, line 125: You're storing it in memory, which isn't enough. In the use case of a user opening their browser for the first time of the day, but using a valid token in localStorage, they should also skip the initialization check.

jonathan-marmelab: I'm not sure that this is a good idea to perist it in the local storage: if a developer restets its DB, the web app will consider the app to be initialized, which is not the case

fzaninotto (approved):

PR #43, opened 2024-08-08
Feat(auth): Handle resetPassword and Invite user

arimet:
## Problem

As a new user, I want to know when I'm invited to join the CRM and configure my own password

## Solution

Implement setPassword and Invite user via Supabase. Handle in HashRouter

### Reset
1. Launch demo
2. From Login page go to ResetPasswordPage. 
3. Send recovery email and go to http://localhost:54324/
4. Click on link and reset password

### Invite user
1. Go on Sales Create
2. Create a new sales and go to http://localhost:54324/
3. Click on mail and set password

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
         "papaparse": "^5.4.1",
         "prop-types": "^15.8.1",
         "ra-data-fakerest": "^5.1.1",
-        "ra-supabase": "^3.0.0",
+        "ra-supabase": "^3.0.1",
         "react": "^18.2.0",
         "react-admin": "^5.1.0",
         "react-cropper": "^2.3.3",
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
         "papaparse": "^5.4.1",
         "prop-types": "^15.8.1",
         "ra-data-fakerest": "^5.1.1",
-        "ra-supabase": "^3.0.1",
+        "ra-supabase": "^3.1.0",
         "react": "^18.2.0",
         "react-admin": "^5.1.0",
         "react-cropper": "^2.3.3",
--- a/src/sales/SalesCreate.tsx
+++ b/src/sales/SalesCreate.tsx
@@ -1,8 +1,6 @@
 import { Card, Container, Typography } from '@mui/material';
 import { useMutation } from '@tanstack/react-query';
 import {
-    PasswordInput,
-    required,
     SimpleForm,
     useDataProvider,
     useNotify,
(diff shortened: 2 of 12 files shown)


jonathan-marmelab (approved):

supabase/functions/users/index.ts
```diff
@@ -35,10 +35,12 @@ async function inviteUser(req: Request) {
         await supabaseAdmin.auth.admin.createUser({
             email,
             password,
-            email_confirm: true,
             user_metadata: { first_name, last_name },
         });
 
+    const { data: emailData, error: emailError } =
```
jonathan-marmelab, line 41: lint error

package.json
```diff
@@ -19,7 +19,7 @@
         "papaparse": "^5.4.1",
         "prop-types": "^15.8.1",
         "ra-data-fakerest": "^5.1.1",
-        "ra-supabase": "^3.0.0",
+        "ra-supabase": "^3.0.1",
```
slax57, line 22: nitpick: there is no 3.0.1. You should require a 3.1.0 instead.


```suggestion
        "ra-supabase": "^3.1.0",
```

supabase/functions/users/index.ts
```diff
@@ -35,11 +35,13 @@ async function inviteUser(req: Request) {
         await supabaseAdmin.auth.admin.createUser({
             email,
             password,
-            email_confirm: true,
             user_metadata: { first_name, last_name },
         });
 
-    if (!data?.user || userError) {
+    const { _, error: emailError } =
```
jonathan-marmelab, line 41: **nit-pick:** The underscore is probably not necessary

```suggestion
    const { error: emailError } =
```

supabase/config.toml
```diff
@@ -71,7 +71,7 @@ enabled = true
 # in emails.
 site_url = "http://localhost:5173/"
 # A list of *exact* URLs that auth providers are permitted to redirect to post authentication.
-additional_redirect_urls = ["https://localhost:5173/auth-callback/index.html"]
+additional_redirect_urls = ["https://localhost:5173/auth-callback.html"]
```
slax57, line 74: don't you need to update the documentation following this change?

supabase/templates/invite.html
```diff
@@ -0,0 +1,90 @@
...
+    <body>
+        <div class="container">
+            <div class="header">
+                <h2>Confirm your account</h2>
+            </div>
+            <div class="content">
+                <p>Welcome to Atomic CRM.</p>
+                <p>
+                    You've been invited to join the (Name) organization. Confirm
+                    your address below:
```
slax57, line 64: This too looks like something that we should document (replacing the organization name in the template)

src/sales/SalesCreate.tsx
```diff
@@ -39,13 +37,7 @@ export function SalesCreate() {
             <Card>
                 <SimpleForm onSubmit={onSubmit as SubmitHandler<any>}>
                     <Typography variant="h6">Create sale person</Typography>
-                    <SalesInputs>
-                        <PasswordInput
-                            source="password"
-                            validate={required()}
-                            helperText={false}
-                        />
```
slax57, line 47: Doesn't this break the fakerest implementation? :thinking:

slax57:
Also, I think I found a small bug.
When you invite a new Admin, after they fill in their new password, they don't see the 'Users' view right away.
They need to sign out / sign in again to see it.

jonathan-marmelab (approved):

PR #44, opened 2024-08-08
fix(crm): Fix signup page logo color

slax57:
## Problem

On the signup page, the logo is almost invisible (white on a white background).

image

## Solution

Invert the logo color on the signup page

## How To Test

Clear your database and run `make start`


jonathan-marmelab (commented):
**question:** Shouldn't the logo be the same blue as the header ?

jonathan-marmelab (approved):

PR #49, opened 2024-08-08
fix(login): Fix user is not automatically logged in after signup

slax57:
## Problem

After sign up, the user is not automatically logged in.

the user has to log in after sign up for the first user

the same happens after I reset my password: I am not automatically logged in.

## Solution

Apply the same `checkAuth` implementation as in [`ra-supabase`](https://github.com/marmelab/ra-supabase/pull/67/files#diff-2f66cc28128c0000e6210759afdbf8eb6cb7597375968d7ab4e3709e2c34cad1), and also extend the mechanism to the signup page.

## How To Test

1st issue: reset the db, create the 1st user, make sure you are automatically signed in.

2nd issue: use the forgot-password feature to reset your password. After choosing a new password, make sure you are automatically signed in.

### DIFF
--- a/src/providers/supabase/authProvider.ts
+++ b/src/providers/supabase/authProvider.ts
@@ -55,6 +55,28 @@ export namespace getIsInitialized {
 export const authProvider: AuthProvider = {
     ...baseAuthProvider,
     checkAuth: async params => {
+        // Users are on the set-password page, nothing to do
+        if (
+            window.location.pathname === '/set-password' ||
+            window.location.hash.includes('#/set-password')
+        ) {
+            return;
+        }
+        // Users are on the forgot-password page, nothing to do
+        if (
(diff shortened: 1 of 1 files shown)


src/providers/supabase/authProvider.ts
```diff
@@ -55,6 +55,28 @@ export namespace getIsInitialized {
 export const authProvider: AuthProvider = {
     ...baseAuthProvider,
     checkAuth: async params => {
+        // Users are on the set-password page, nothing to do
```
fzaninotto, line 58: Why is checkAuth called on these pages?

slax57: TBH I don't know, but I figured applying the same implementation as what ra-supabase does couldn't be a bad thing.

PR #61, opened 2024-08-12
Feat(tasks): Update contact last seen when creating a task

jonathan-marmelab:
## Problem

Contact last activity is not updated when a task has been added

## Solution

Add an onSucces hook that updates the contact last activity when creating a task

## How To Test

Add a task to a contact

## Additional Checks

### DIFF
--- a/src/tasks/AddTask.tsx
+++ b/src/tasks/AddTask.tsx
@@ -27,6 +27,7 @@ import {
     useGetIdentity,
     useNotify,
     useRecordContext,
+    useUpdate,
 } from 'react-admin';
 import { Link } from 'react-router-dom';
 import { contactInputText, contactOptionText } from '../misc/ContactOption';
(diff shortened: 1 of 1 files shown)


src/tasks/AddTask.tsx
```diff
@@ -40,13 +41,30 @@ export const AddTask = ({
...
+        const contact = await dataProvider.getOne('contacts', {
+            id: data.contact_id,
+        });
+        if (!contact.data) return;
+
+        dataProvider.update('contacts', {
+            id: contact.data.id,
+            data: { last_seen: new Date().toISOString() },
+            previousData: contact.data,
+        });
```
slax57, line 64: **Suggestions:**

- You should `await` the mutaiton before displaying the success notification
- You should use `useUpdate` from `react-admin` instead of calling the dataProvider directly, as this will allow to update the react-query cache, so that the data is up to date without having to wait for the refresh on the ContactList page

slax57 (approved):

PR #62, opened 2024-08-12
Feat/mail company name

jonathan-marmelab:
## Problem

- The company name is the company domain name, which is not what the user expects

## Solution

- Improve company name parsing from domain

## How To Test

- Send a mail with inbound e-mail address

## Additional Checks

- [ ] ~~Tested with **fakerest** provider (see [related documentation](../doc/data-providers.md))~~ *(N/A)*

### DIFF
--- a/public/auth-callback.html
+++ b/public/auth-callback.html
@@ -131,7 +131,7 @@
         function interceptAuthCallback() {
                 const hash = window.location.hash;
                 const {access_token, refresh_token, type} = getUrlParams();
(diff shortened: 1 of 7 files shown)


slax57 (changes requested):

jonathan-marmelab:
Setting up a library leads to problems with unit tests due to the incompatibility with node and Deno. We will need to invest on this to set up everything correctly

PR #63, opened 2024-08-13
Feat(task): Add task edit support

jonathan-marmelab:
## Problem

It is not possible to edit a tag

## Solution

Add tag edit support

## How To Test

Create a tag and edit it

## Additional Checks

### DIFF
--- a/src/tags/TagChip.tsx
+++ b/src/tags/TagChip.tsx
@@ -1,0 +1,36 @@
+import { Chip } from '@mui/material';
+import { useState } from 'react';
+import { Tag } from '../types';
+import { TagEditModal } from './TagEditModal';
+
+type TagChipProps = {
+    tag: Tag;
+
+    onUnlink: () => Promise<void>;
+};
+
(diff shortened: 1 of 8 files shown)


src/tags/TagChip.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Tag } from '../types';
+import { TagEditModal } from './TagEditModal';
+
+type TagChipProps = {
+    tag: Tag;
+
+    onUnlink: () => Promise<void>;
+};
+
+export function TagChip({ tag, onUnlink }: TagChipProps) {
```
slax57, line 12: **question**: why call that `onUnlink`? :thinking:

slax57 (approved):

src/tags/TagChip.tsx
jonathan-marmelab: I did not want to call this onDelete, since delete naming will be used for deletion once it is implemented

PR #66, opened 2024-09-13
Add a build-lib command to be able to publish an atomic-crm node module

ThieryMichel:
## Problem

We cannot install atomic-crm from npm

## Solution

Add a lib build along the current app build with:
- esm 
- cjs
- types

## How To Test

To test without actually publishing the module:
- run npm pack to create an archive
- create a npm project, aor use an existing one
- add `"atomic-crm": "file://path-to-the-archive/atomic-crm-0.0.1.tgc"`
- run npm install
- start to play with the atomic-crm library
- Admire the type
- switch your project type between module and commonjs to use both version.
Note: only the CRM component is exposed for now 

## Additional Checks

- [ ] Tested with **fakerest** provider (see [related documentation](../doc/data-providers.md))

### DIFF
--- a/makefile
+++ b/makefile
@@ -25,6 +25,9 @@ stop: stop-supabase ## stop the stack locally
 build: ## build the app
 	npm run build
 
+build-lib: ## build the library
+	npm run build-lib
+
 prod-start: build supabase-deploy
 	open http://127.0.0.1:3000 && npx serve -l tcp://127.0.0.1:3000 dist
 
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,18 +1,3 @@
-import React from 'react';
-import { createRoot } from 'react-dom/client';
-import App from './App';
-import reportWebVitals from './reportWebVitals';
+import { CRM } from './root/CRM';
 
-const container = document.getElementById('root');
-const root = createRoot(container!);
-
-root.render(
-    <React.StrictMode>
-        <App />
-    </React.StrictMode>
-);
-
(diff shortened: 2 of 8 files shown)


src/index.tsx
```diff
@@ -1,18 +1,3 @@
...
-    <React.StrictMode>
-        <App />
-    </React.StrictMode>
-);
-
-// If you want to start measuring performance in your app, pass a function
-// to log results (for example: reportWebVitals(console.log))
-// or send to an analytics endpoint. Learn more: https://bit.ly/CRA-vitals
-reportWebVitals();
+export default CRM;
```
fzaninotto, line 3: **Nitcpick**: I'd prefer that we don't use the default export, in order to be able to export more than one component without breaking apps that used the first version of the package.

fzaninotto (changes requested):
Excellent, thanks for that!

fzaninotto (approved):

makefile
```diff
@@ -25,6 +25,9 @@ stop: stop-supabase ## stop the stack locally
 build: ## build the app
 	npm run build
 
+build-lib: ## build the library
+	npm run build-lib
+
```
slax57, line 30: Shouldn't this be added to the `make build` target? I'm worried we risk forgetting to run `make build-lib`  next time we release the package to npm :grimacing:  .

slax57: Actually both builds are separate: one is used to deploy the CRM app, the other is used to build the lib for npm. So keeping the 2 targets separate is a good thing.

PR #70, opened 2024-10-24
Access Control: use canAccess instead of permissions

djhi:
## Problem

React Admin `5.3.0` introduced a better [access control mechanism](https://marmelab.com/react-admin/Permissions.html) but Atomic CRM still uses the old permissions.

## Solution

Convert the usage of permissions to canAccess.

## How To Test

- Follow the initial setup from the README

### As an admin (the default)
- Check that you can access the users admin menu
- Check that you can access the sales pages
- Note an email of a "simple" sale

### As the "simple" sale
- There is no *Users* menu item in the top right user menu dropdown
- Trying to access any sales URL by typing it directly in the browser address bar should redirect to the `/access-denied` page

## Additional Checks

- [x] Tested with **fakerest** provider (see [related documentation](../doc/data-providers.md))


src/providers/fakerest/authProvider.ts
```diff
@@ -50,19 +51,21 @@ export const authProvider: AuthProvider = {
...
         const localUser = userItem ? (JSON.parse(userItem) as Sale) : null;
-        if (!localUser) {
-            return Promise.reject('user is not logged in');
-        }
+        if (!localUser) return false;
 
-        // We fetch permissions from server to avoid local storage tampering
-        const user = await dataProvider.getOne<Sale>('sales', {
+        // Get the matching sale
+        const { data } = await dataProvider.getOne<Sale>('sales', {
```
slax57, line 61: same as https://github.com/marmelab/react-admin/pull/10300#discussion_r1816783310

src/providers/supabase/authProvider.ts
```diff
@@ -90,8 +83,26 @@ export const authProvider: AuthProvider = {
...
-        return isInitialized ? baseAuthProvider.getPermissions(params) : null;
+        if (!isInitialized) return false;
+
+        // Get the current user
+        const { data: dataSession, error: errorSession } =
+            await supabase.auth.getSession();
+        if (!dataSession.session?.user || errorSession) return false;
+
+        // Get the matching sale
+        const { data: dataSale, error: errorSale } = await supabase
```
slax57, line 97: same

slax57 (changes requested):

src/providers/supabase/authProvider.ts
```diff
@@ -90,8 +83,26 @@ export const authProvider: AuthProvider = {
 
         return baseAuthProvider.checkAuth(params);
     },
-    async getPermissions(params) {
+    // FIXME: supabaseAuthProvider does not yet provide canAccess
```
fzaninotto, line 86: I don't see how we can provide a generic `canAccess` in `ra-supabase`, so this FIXME should be removed IMHO

djhi: Like we do for `getPermissions` and `getIdentity`?

src/providers/supabase/authProvider.ts
```diff
@@ -20,20 +21,12 @@ const baseAuthProvider = supabaseAuthProvider(supabase, {
...
-            .single();
-
-        if (!data || error) {
-            return null;
-        }
-
-        return data?.administrator ? 'admin' : 'user';
-    },
 });
+// FIXME: Now that react-admin pessimistically calls getPermissions, it calls getPermissions when it initializes its routes
```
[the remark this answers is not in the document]
slax57: Actually this is because of the default implementation of `getPermissions` in `ra-supabase`, which throws when the user is not authenticated.
Should be fixed by https://github.com/marmelab/ra-supabase/pull/79.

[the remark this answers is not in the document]
fzaninotto: That's indeed a bug, but not a regression : it comes from ra-supabase. We will be able to remove this code once https://github.com/marmelab/ra-supabase/pull/79 is merged and published.

src/providers/supabase/authProvider.ts
```diff
@@ -1,39 +1,28 @@
...
 
 const baseAuthProvider = supabaseAuthProvider(supabase, {
-    getIdentity: async user => {
-        const { data, error } = await supabase
-            .from('sales')
-            .select('id, first_name, last_name, avatar')
-            .match({ user_id: user.id })
-            .single();
+    getIdentity: async () => {
+        const sale = getSaleFromLocalStorage();
```
slax57, line 8: I know I'm the one who asked to cache the connected `sale` to avoid multiple calls to supabase when calling `canAccess`, but this new implementation causes an issue: when you invite a new user, click on the invitation link in the email, and fill in your password, then you are still connected as the admin you were before, and not as the newly created sales.

Maybe the easiest fix is to use a local variable instead of the localStorage?

slax57 (changes requested):

src/providers/supabase/authProvider.ts
slax57: fixed

slax57 (approved):

PR #74, opened 2024-11-04
Add more secure RLS policies on the sales table

djhi:
## Problem

The RLS policies on the `sales` table are very permissive. Any authenticated users can give themselves the administrator privilege.

## Solution

- [x] Add stricter policies:
  - Only admins can add new users
  - Only admins can delete users
  - Nobody can update sales directly
- [x] Fix supabase functions so that simple users can only update their non security related data
 
## How To Test

- Use the dev console to copy a user token and the api key from a postgrest call as a simple authenticated user
- Use a tool such as Bruno to send an authenticated `POST` request to the users endpoint (`http://127.0.0.1:54321/functions/v1/users`), trying to set the `administrator` and/or `disabled` columns to `true`. It should ignore security related changes. Example request body:
```json
{
  "sales_id": 2,
  "first_name": "Toto",
  "last_name": "Totooooo",
  "email": "toto@marmelab.com",
  "administrator": true       ,
  "disabled": true
}
```
- Use a tool such as Bruno to send an authenticated `POST` request to the users endpoint (`http://127.0.0.1:54321/functions/v1/users`), trying to update another user (change the `sales_id`). It should fail with a 401.
- As this user, verify you can still update your own data either with Bruno or the admin UI
- As an admin, verify you can still update any user and change their security related options (admin, disabled)
- Check that no one can make PATCH requests to the `sales` endpoint (e.g. `http://127.0.0.1:54321/rest/v1/sales?id=eq.2`) using Bruno.


erwanMarmelab (approved):

fzaninotto:
It seems the ability to change the avatar for a sale is broken. Not sure if it's related to this  change though. 

Apart from that, my tests are OK.

supabase/functions/users/index.ts
```diff
@@ -103,6 +124,10 @@ async function patchUser(req: Request, currentUserSale: any) {
         return createErrorResponse(500, 'Internal Server Error');
     }
 
+    if (avatar) {
+        await updateSaleAvatar(data.user.id, avatar);
```
fzaninotto, line 128: Something is wrong. This edge function is still accessible to anonymous users using the [remote function API](https://supabase.com/docs/guides/functions/deploy#invoking-remote-functions) right?

djhi: It does, see:
- https://github.com/marmelab/atomic-crm/pull/74/files/b82310e675e8bb18b51169e671f5c45b78426dd0..2c2ed26627070fa5c34390eb25f10beca50467a1#diff-1d0b413d00a6df280ee14f49c3f5766d5ac5bd57ad21fdb338334bc15c3093baR111
- https://github.com/marmelab/atomic-crm/pull/74/files#diff-1d0b413d00a6df280ee14f49c3f5766d5ac5bd57ad21fdb338334bc15c3093baR189-R197

fzaninotto: My bad, you're right!

fzaninotto (approved):

fzaninotto:
needs rebase

PR #79, opened 2025-01-08
Add telemetry

djhi:
## Problem

We have no idea of how many developers actually use Atomic CRM in production.

## Solution

Add telemetry like in React Admin.

### DIFF
--- a/src/root/CRM.tsx
+++ b/src/root/CRM.tsx
@@ -1,3 +1,4 @@
+import { useEffect } from 'react';
 import {
     Admin,
     CustomRoutes,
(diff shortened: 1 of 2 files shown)


src/root/CRM.tsx
```diff
@@ -127,55 +128,76 @@ export const CRM = ({
...
+            disableTelemetry ||
+            process.env.NODE_ENV !== 'production' ||
+            typeof window === 'undefined' ||
+            typeof window.location === 'undefined' ||
+            typeof Image === 'undefined'
+        ) {
+            return;
+        }
+        const img = new Image();
+        img.src = `//https://atomic-crm-telemetry.marmelab.com/atomic-crm-telemetry?domain=${window.location.hostname}`;
```
djhi, line 145:
```suggestion
        img.src = `https://atomic-crm-telemetry.marmelab.com/atomic-crm-telemetry?domain=${window.location.hostname}`;
```

slax57 (approved):

PR #95, opened 2025-03-28
Make the reset command apply the last migrations

erwanMarmelab:
## Problem

When I reset my database, the last migrations are not applied. 

## Solution

Use the migration command on database reset

### DIFF
--- a/makefile
+++ b/makefile
@@ -16,8 +16,8 @@ supabase-migrate-database: ## apply the migrations to the database
 	npx supabase migration up
 
 supabase-reset-database: ## reset (and clear!) the database
-	$(MAKE) supabase-migrate-database
 	npx supabase db reset
+	$(MAKE) supabase-migrate-database
 
 start-app: ## start the app locally
 	npm run dev
--- a/makefile
+++ b/makefile
@@ -16,6 +16,7 @@ supabase-migrate-database: ## apply the migrations to the database
 	npx supabase migration up
 
 supabase-reset-database: ## reset (and clear!) the database
+	$(MAKE) supabase-migrate-database
 	npx supabase db reset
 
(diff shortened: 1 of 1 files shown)


makefile
```diff
@@ -16,6 +16,7 @@ supabase-migrate-database: ## apply the migrations to the database
 	npx supabase migration up
 
 supabase-reset-database: ## reset (and clear!) the database
+	$(MAKE) supabase-migrate-database
```
slax57, line 19: shouldn't you inverse the order? :thinking:

makefile
```diff
@@ -16,6 +16,7 @@ supabase-migrate-database: ## apply the migrations to the database
...
 	npx supabase db reset
```
erwanMarmelab, line 20:
```suggestion
	npx supabase db reset
	$(MAKE) supabase-migrate-database
```

makefile
erwanMarmelab: why not, both works. Applied

makefile
```diff
@@ -17,6 +17,7 @@ supabase-migrate-database: ## apply the migrations to the database
 
 supabase-reset-database: ## reset (and clear!) the database
 	npx supabase db reset
+	$(MAKE) supabase-migrate-database
```
slax57, line 20: I still have doubts this is really needed. 
According the the [docs](https://supabase.com/docs/reference/cli/supabase-db-reset), it should already apply migrations at this stage:

> Recreates the local Postgres container and applies all local migrations found in supabase/migrations directory

Can you double check?

slax57 (changes requested):

makefile
djhi: This shouldn't be needed. The `supabase db reset` command apply all migrations

PR #104, opened 2025-09-02
Replace react admin with shadcn admin kit

Madeorsk:
The main Atomic CRM version is still the one based on React Admin. Replace it with our version of Atomic CRM based on shadcn admin kit.

## TODO

- [x] The contact list doesn’t have a bulk actions toolbar. Selecting items in the list has no effect.
- [x] In the contact show view, adding a note (without showing options) doesn’t set the contact status. In the previous version, every note has a status, whether the user has set it or not.
- [x] In the contact show view, clicking on the “show options” link under the new note form reveals the options without animation. There used to be a sliding reveal.
- [x] When updating a note, the “update note” button should be primary.
- [x] In the tags list on the sidebar of  a contact show, the line height between tags is too high. The vertical gap should be roughjly the same as the horizontal one.
- [x] The default colors for the tags are ugly. Can’t we reuse the same as in the previous version ?
- [x] The company list isn’t as responsive as the old one, which had a variable number of columns depending on the window size
- [ ] ~In a company edit view, changing the company avatar by zooming and panning seems to work in optimistic, but once the company reloads, the avatar isn’t correct~ (previously existing bug)
- [x] In the deals list, the filters are on a new line compared to the actions toolbar, which moves the content down
- [x] The deals list shows a pagination, there shouldn’t be one.
- [x] In the deal creation dialog, it used to be possible to create a company on-the fly in the company selector.
- [x] When editing a deal, the font of the dialog title changes (so does the buttons toolbar)
- [x] In the “my info” page, a long inbound email breaks free from its container. Also, this container looks like an input, which is wrong - the user can’t update the inbound email.
- [x] In the sales list, the “Sales” title isn’t necessary. Also, put the filters and the actions in the same row.

## Differences with atomic-crm-shadcn

- Update ra-supabase-core
- Add missing pages and configuration for supabase
- Fix signup page spacing
- Setup fakerest demo and improve welcome card
- Fix sales and contact lists
- Fix text input when a date is provided
- Create a company from autocomplete input
- Update shadcn-admin-kit to the latest version
- Add set password / forgot password pages

## How To Test

- Test with supabase: `make start`
- Test in demo mode: `make start-demo`
- Build and preview demo mode: `make build-demo && npm run preview`
- Build and preview supabase: `make build && npm run preview`

## Additional Checks

- [X] Tested with **fakerest** provider (see [related documentation](https://github.com/marmelab/atomic-crm/blob/main/doc/developer/data-providers.md))


demo/main.tsx
```diff
@@ -0,0 +1,11 @@
+import { StrictMode } from "react";
+import { createRoot } from "react-dom/client";
+import "../src/index.css";
+import App from "./App.tsx";
+
+// biome-ignore lint/style/noNonNullAssertion: #root element is guaranteed to exist
```
djhi: Why do we have biome related comments when we use eslint?

src/App.tsx
```diff
@@ -1,30 +1,5 @@
...
- * ... as well as all the props accepted by react-admin's <Admin> component.
- *
- * @example
- * const App = () => (
- *    <CRM
- *       logo="./img/logo.png"
- *       title="Acme CRM"
- *    />
- * );
- */
```
djhi: Not sure why we removed this. It's seems still relevant and useful

src/atomic-crm/activity/ActivityLog.tsx
```diff
@@ -0,0 +1,54 @@
+import { Alert } from "@/components/ui/alert";
+import { Separator } from "@/components/ui/separator";
+import { Skeleton } from "@/components/ui/skeleton";
+import { useQuery } from "@tanstack/react-query";
+import { Identifier, useDataProvider } from "ra-core";
```
djhi:
```suggestion
import { type Identifier, useDataProvider } from "ra-core";
```

src/atomic-crm/activity/ActivityLog.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import { CrmDataProvider } from "../providers/types";
```
djhi:
```suggestion
import type { CrmDataProvider } from "../providers/types";
```

src/atomic-crm/activity/ActivityLogDealCreated.tsx
```diff
@@ -0,0 +1,41 @@
+import { Link } from "react-router";
+
+import { RaRecord } from "ra-core";
```
djhi:
```suggestion
import type { RaRecord } from "ra-core";
```

src/atomic-crm/activity/ActivityLogDealNoteCreated.tsx
```diff
@@ -0,0 +1,84 @@
+import { ReferenceField } from "@/components/admin";
+
+import { RaRecord } from "ra-core";
```
djhi:
```suggestion
import type { RaRecord } from "ra-core";
```

src/atomic-crm/activity/ActivityLogIterator.tsx
```diff
@@ -0,0 +1,79 @@
+import { Separator } from "@/components/ui/separator";
+import { Fragment, useState } from "react";
+import {
+  COMPANY_CREATED,
+  CONTACT_CREATED,
+  CONTACT_NOTE_CREATED,
+  DEAL_CREATED,
+  DEAL_NOTE_CREATED,
+} from "../consts";
+import { Activity } from "../types";
```
djhi:
```suggestion
import type { Activity } from "../types";
```

djhi (changes requested):
Only reviewed a small part as the GH UI is lagging a lot:
- We're importing a lot of types without the `type` prefix
- Didn't we have the ability to set a contact status before (cold, warm, etc) when adding  a note?

Issue #106, opened 2025-09-29
UI Revamp: Contact Sidebar with Search Input, Icon Contact Filter list


mpsalunggg:
I found that the placeholder text overlaps with the search icon in the contact input field.
Additionally, the contact sidebar icons are inconsistent in their width and height.

I’d like to contribute a small fix for this issue. I’m still learning, but I hope this contribution can be helpful.

This is the ui issues
- Input Search
<img width="454" height="161" alt="Image" src="https://github.com/user-attachments/assets/7527549e-8204-43e4-b0e0-a9b21efb71e9" />

- Icon list filter
<img width="278" height="206" alt="Image" src="https://github.com/user-attachments/assets/9b021a3a-ee92-4736-8fdb-8c2ad23efc6f" />

slax57:
Indeed! Thanks for the report and for offering to work on a fix! 🙂

PR #107, opened 2025-09-29
[CHORE] - adjust UI contact component search input, icon contact filter

mpsalunggg:
Closes https://github.com/marmelab/atomic-crm/issues/106

## Problem

I found that the placeholder text overlaps with the search icon in the contact input field.
Additionally, the contact sidebar icons are inconsistent in their width and height.
- Search Input
Screenshot 2025-09-29 at 14 55 28

- Icon Contact Filter
Screenshot 2025-09-29 at 15 42 43

## Solution

* Changed `w-fit` to `w-full` in the `search-input` component to prevent the placeholder from overlapping with the search icon.
* Removed fixed `width` and `height` attributes on the sidebar contact icons to ensure consistent sizing with the existing icons.

## How To Test

1. Navigate to the **Contacts** section.
2. Verify that the search input placeholder no longer overlaps with the search icon.
3. Check the contact sidebar icons to confirm that their width and height are now consistent with each other.

## Additional Checks

* [ ] The **documentation** is up to date
* [ ] Tested with **fakerest** provider (see [[related documentation](https://github.com/marmelab/atomic-crm/blob/main/doc/developer/data-providers.md)](https://github.com/marmelab/atomic-crm/blob/main/doc/developer/data-providers.md))

Also, please make sure to read the [[contributing guidelines](https://github.com/marmelab/atomic-crm#contributing)](https://github.com/marmelab/atomic-crm#contributing).

### DIFF
--- a/src/components/admin/search-input.tsx
+++ b/src/components/admin/search-input.tsx
@@ -14,7 +14,7 @@ export const SearchInput = (inProps: SearchInputProps) => {
   }
 
   return (
-    <div className="flex flex-grow relative mt-auto w-full">
+    <div className="flex flex-grow relative mt-auto">
       <TextInput
         label={false}
         helperText={false}
--- a/src/components/admin/search-input.tsx
+++ b/src/components/admin/search-input.tsx
@@ -14,7 +14,7 @@ export const SearchInput = (inProps: SearchInputProps) => {
   }
 
   return (
-    <div className="flex flex-grow relative mt-auto w-fit">
+    <div className="flex flex-grow relative mt-auto w-full">
       <TextInput
         label={false}
         helperText={false}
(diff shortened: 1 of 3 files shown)


src/components/admin/search-input.tsx
```diff
@@ -14,7 +14,7 @@ export const SearchInput = (inProps: SearchInputProps) => {
   }
 
   return (
-    <div className="flex flex-grow relative mt-auto w-fit">
+    <div className="flex flex-grow relative mt-auto w-full">
```
slax57: This is probably not the correct fix. If the container is wider, then the icon will end up outside of the input.

image


We should probably instead fix the width of the TextInput

mpsalunggg: Hmm, when I run it locally and use w-full, it looks like this
image
The icon is not missing from the input 🤔

slax57: Indeed in the context of this specific filter sidebar it works, but I wanted the fix to be more generic.
Thus I applied the same fix as https://github.com/marmelab/shadcn-admin-kit/pull/64

slax57 (approved):
Thanks in any case!

PR #115, opened 2025-10-24
Chore: Add registry file for Atomic CRM

jonathanarnault:
## Problem

The Atomic CRM components keep evolving. 
Yet it's hard to update an existing installation of Atomic CRM, as the installation instructions advise to forl the repo. 

## Solution

Add a registry for Atomic CRM, so that users may update their Atomic CRM components with the `shadcn` command. 

## How To Test

```bash
make registry-build
````

Then create a new shadcn project and then run

```bash
npx shadcn add http://localhost:5173/r/atomic-crm.json -o
```

## Additional Checks

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,6 @@
     "husky": "^9.1.7",
     "prettier": "^3.6.2",
     "shadcn": "^3.5.0",
-    "supabase": "^2.53.6",
     "typescript": "~5.8.3",
     "typescript-eslint": "^8.35.1",
     "vite": "^7.0.4",
--- a/package-lock.json
+++ b/package-lock.json
@@ -82,7 +82,6 @@
         "husky": "^9.1.7",
         "prettier": "^3.6.2",
         "shadcn": "^3.5.0",
-        "supabase": "^2.53.6",
         "typescript": "~5.8.3",
         "typescript-eslint": "^8.35.1",
         "vite": "^7.0.4",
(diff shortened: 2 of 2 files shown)


src/atomic-crm/activity/ActivityLog.tsx
```diff
@@ -2,12 +2,10 @@ import { Alert } from "@/components/ui/alert";
 import { Separator } from "@/components/ui/separator";
 import { Skeleton } from "@/components/ui/skeleton";
 import { useQuery } from "@tanstack/react-query";
-import type { Identifier } from "ra-core";
-import { useDataProvider } from "ra-core";
-
-import type { CrmDataProvider } from "../providers/types";
```
fzaninotto, line 8: The interest of import aliases is to make the import path shorter and not care about relative position. It's relevant if all elements are all in the same directtoy, which isn't the case with atomic CRM as there are too many files and we kept a hierarchy. 

So I think we should not change the relative import paths to aliases for atomic CRM.

makefile
```diff
@@ -84,4 +84,14 @@ doc-preview: doc-build
...
-	@(cd doc && npx gh-pages -d dist -e doc -b gh-pages -m "Deploy docs" --remove doc)
+	@(cd doc && npx gh-pages -b gh-pages -d dist -e doc -m "Deploy docs" --remove doc)
+
+registry-build: ## build the shadcn registry
+	npm run registry:build
+
+registry-deploy: registry-build ## deploy the shadcn registry
+	@(cd public/r && npx gh-pages -b gh-pages -s atomic-crm.json -e r -m "Deploy registry" --remove r)
+
+registry-gen: ## generate the shadcn registry
```
fzaninotto: explain in a comment that it's called in a pre-commit hook, it shouldn't be called by hand

makefile
```diff
@@ -84,4 +84,14 @@ doc-preview: doc-build
 	@(cd doc && npm run preview)
 
 doc-deploy:
-	@(cd doc && npx gh-pages -d dist -e doc -b gh-pages -m "Deploy docs" --remove doc)
+	@(cd doc && npx gh-pages -b gh-pages -d dist -e doc -m "Deploy docs" --remove doc)
+
+registry-build: ## build the shadcn registry
```
fzaninotto, line 89: explain in a comment that it's called in a CI pipeline, it shouldn't be called by hand

src/atomic-crm/activity/ActivityLog.tsx
```diff
@@ -2,9 +2,7 @@ import { Alert } from "@/components/ui/alert";
 import { Separator } from "@/components/ui/separator";
 import { Skeleton } from "@/components/ui/skeleton";
 import { useQuery } from "@tanstack/react-query";
-import type { Identifier } from "ra-core";
-import { useDataProvider } from "ra-core";
-
```
fzaninotto, line 7: The extra newline between external imports and internal imports is on purpose (for legibility). Please revert its removal here and in other files.

package.json
```diff
@@ -86,8 +88,13 @@
     "eslint-plugin-react-hooks": "^5.2.0",
     "eslint-plugin-react-refresh": "^0.4.19",
     "eslint-plugin-storybook": "^9.0.13",
+    "gh-pages": "^6.3.0",
+    "glob": "^11.0.3",
     "globals": "^16.3.0",
+    "husky": "^9.1.7",
     "prettier": "^3.6.2",
+    "shadcn": "^3.5.0",
+    "supabase": "^2.53.6",
```
fzaninotto: Why did you need to add this one?

fzaninotto (approved):

PR #132, opened 2026-01-06
Fix signup error notification not being displayed

WiXSL:
## Problem
SignupPage component don't show notifications on error

## How to Test

Creating the first user, leave the password input with one character and submit the form.
Expected: A notification is shown with the error.
Actual: Nothing is shown to the user and the form is not submited

## Solution

Add a `Notification` component to the markup

### DIFF
--- a/src/components/atomic-crm/login/SignupPage.tsx
+++ b/src/components/atomic-crm/login/SignupPage.tsx
@@ -42,7 +42,7 @@ export const SignupPage = () => {
         });
       });
     },
-    onError: error => {
+    onError: (error) => {
       notify(error.message);
     },
   });


src/components/atomic-crm/login/SignupPage.tsx
```diff
@@ -11,6 +11,7 @@ import type { CrmDataProvider } from "../providers/types";
 import { useConfigurationContext } from "../root/ConfigurationContext";
 import type { SignUpData } from "../types";
 import { LoginSkeleton } from "./LoginSkeleton";
+import { Notification } from "@/components/admin";
```
fzaninotto: We use direct import internally in atomic crm

```suggestion
import { Notification } from "@/components/admin/notification";
```

PR #133, opened 2026-01-06
Allow to import data from another CRM

djhi:
## Problem

Some users want to migrate from other CRMs to AtomicCRM

## Solution

Implement an importer that takes a generic JSON file containing contacts, companies, tasks and notes. Attachments are URLs in the JSON file to publicly accessible resources. 

- [x] Initial implementation with Fakerest (slow)
- [x] Make it work with Supabase
- [x] Handle contacts avatar with an edge function and a trigger
- [x] Handle note import with an edge function that would take care of attachments
~~- [ ] Implement a `createMany` dataProvider function and refactor batching accordingly~~

### Sample import files

- [With only valid data](https://github.com/user-attachments/files/24584825/test-highrise.json)
- [With invalid data](https://github.com/user-attachments/files/24584830/test-invalid-highrise.json)

When using the _With invalid data_ file, check that the import dialog offers you to download an error report.

### FakeRest

Import is particularly slow with FakeRest as we have to update avatars client side.
- `make start-demo`
- Click the user menu and the _Import from Highrise_ item
- Select a file generated from the exporter

### Supabase

- `make start`
- Click the user menu and the _Import from Highrise_ item
- Select a file generated from the exporter

## Additional Checks

### DIFF
--- a/supabase/functions/users/index.ts
+++ b/supabase/functions/users/index.ts
@@ -103,18 +103,14 @@ async function inviteUser(req: Request, currentUserSale: any) {
         .select("*")
         .eq("user_id", user.id);
       if (salesError) {
-        return createErrorResponse(
-          salesError.status,
-          salesError.message,
-          {
-            code: salesError.code,
-          },
-        );
+        return createErrorResponse(salesError.status, salesError.message, {
+          code: salesError.code,
+        });
       }
       if (existingSale.length > 0) {
         return createErrorResponse(
           400,
-          'A sales for this email already exists',
+          "A sales for this email already exists",
         );
       }
 
@@ -147,13 +143,9 @@ async function inviteUser(req: Request, currentUserSale: any) {
   } else {
     if (userError) {
       console.error(`Error inviting user: user_error=${userError}`);
-      return createErrorResponse(
-        userError.status,
-        userError.message,
-        {
-          code: userError.code,
-        },
-      );
+      return createErrorResponse(userError.status, userError.message, {
+        code: userError.code,
+      });
     }
     if (!data?.user) {
       console.error("Error inviting user: undefined user");


src/components/atomic-crm/misc/useImportFromJson.ts
```diff
@@ -0,0 +1,687 @@
...
+        return;
+      }
+      try {
+        const existingRecord = await dataProvider.getList("sales", {
+          filter: { email: dataToImport.email.trim() },
+          pagination: { page: 1, perPage: 1 },
+          sort: { field: "id", order: "ASC" },
+        });
+        if (existingRecord.total === 1) {
+          return existingRecord.data[0].id;
```
WiXSL: When a sale already exists, you return the id but never store it in `idsMaps.sales`, so imported companies/contacts that reference that `sales_id` end up with `sales_id`: undefined.

src/components/atomic-crm/misc/useImportFromJson.ts
```diff
@@ -0,0 +1,687 @@
...
+          break;
+        }
+        case "tasks": {
+          currentBatch.push(importTask(value));
+          break;
+        }
+      }
+      await proccesBatchIfPossible();
+      // Error state should stop the import process
+      if (state.status === "error") {
```
WiXSL: The loop checks state.status === "error" but `state` is stale inside the async loop; even after setState sets error, the loop won’t see it and keeps importing.

WiXSL (changes requested):

src/components/atomic-crm/misc/useImportFromJson.ts
```diff
@@ -0,0 +1,687 @@
...
+      if (!isSale(dataToImport)) {
+        setState((old) => ({
+          ...old,
+          status: "error",
+          error: new Error(`Error while importing sale: Invalid data`),
+          failedImports: {
+            ...old.failedImports,
+            sales: [...old.failedImports.sales, dataToImport],
+          },
+          duration: new Date().getDate() - startedAt.getDate(),
```
WiXSL: Maybe: `Date().valueOf() - startedAt.valueOf()` like in the other cases in this file (?)

src/components/atomic-crm/providers/supabase/dataProvider.ts
```diff
@@ -370,15 +351,32 @@ const uploadToBucket = async (fi: RAFile) => {
...
-        return;
+        return fi;
       }
     }
   }
 
   const dataContent = fi.src
-    ? await fetch(fi.src).then((res) => res.blob())
+    ? await fetch(fi.src, {
+        mode: "no-cors",
```
WiXSL: fetch(..., { mode: "no-cors" }) yields an opaque response with status === 0, so you always treat it as failure and skip uploading attachments.

src/components/atomic-crm/layout/Header.tsx
```diff
@@ -27,64 +31,73 @@ const Header = () => {
...
+                  className="[.light]:hidden h-6"
+                  src={darkModeLogo}
+                  alt={title}
                 />
-                <NavigationTab
-                  label="Companies"
-                  to="/companies"
-                  isActive={currentPath === "/companies"}
+                <img
+                  className="[.dark]:hidden h-6"
```
WiXSL:
```suggestion
                  className="[.dark_&]:hidden h-6"
```

src/components/atomic-crm/layout/Header.tsx
```diff
@@ -27,64 +31,73 @@ const Header = () => {
...
+      <nav className="grow">
+        <header className="bg-secondary">
+          <div className="px-4">
+            <div className="flex justify-between items-center flex-1">
+              <Link
+                to="/"
+                className="flex items-center gap-2 text-secondary-foreground no-underline"
+              >
+                <img
+                  className="[.light]:hidden h-6"
```
WiXSL:
```suggestion
                  className="[.light_&]:hidden h-6"
```

src/components/atomic-crm/layout/Header.tsx
```diff
@@ -27,64 +31,73 @@ const Header = () => {
...
+                </UserMenu>
+              </div>
             </div>
           </div>
-        </div>
-      </header>
-    </nav>
+        </header>
+      </nav>
+      <ImportFromJsonDialog
```
fzaninotto, line 97: I think this dialog and the corresponding state should be part of the `ImportFromJsonButton` for better encapsulation.

src/components/atomic-crm/layout/Header.tsx
```diff
@@ -131,4 +144,19 @@ const ConfigurationMenu = () => {
     </DropdownMenuItem>
   );
 };
+
+const ImportFromJsonButton = ({ onClick }: { onClick: () => void }) => {
+  const { onClose } = useUserMenu() ?? { onClose: () => {} };
```
fzaninotto: the `?? { onClose: () => {} }` condition will hide feailures. Instead, the component should throw if not used inside a user menu.

src/components/atomic-crm/layout/Header.tsx
```diff
@@ -131,4 +144,19 @@ const ConfigurationMenu = () => {
     </DropdownMenuItem>
   );
 };
+
+const ImportFromJsonButton = ({ onClick }: { onClick: () => void }) => {
```
fzaninotto: I find the name confusing since `ImportFromJsonButton` already exists somewhere else. Why not use `ImportFromJsonMenuItem` instead?

src/components/atomic-crm/misc/ImportFromJsonDialog.tsx
```diff
@@ -0,0 +1,232 @@
...
+  onOpenChange,
+  ...props
+}: React.ComponentProps<typeof Dialog>) => {
+  const [importStatus, importFile, reset] = useImportFromJson();
+
+  return (
+    <Dialog
+      onOpenChange={(open) => {
+        if (importStatus.status !== "importing") {
+          reset();
```
fzaninotto, line 46: Can you add a comment to explain why you do that?

src/components/atomic-crm/misc/ImportFromJsonDialog.tsx
```diff
@@ -0,0 +1,232 @@
...
+          }
+        }
+      }}
+      {...props}
+    >
+      <DialogContent>
+        <DialogHeader>
+          <DialogTitle>Import from JSON</DialogTitle>
+        </DialogHeader>
+        {importStatus.status === "idle" || importStatus.status === "error" ? (
```
fzaninotto: I find the ternaries hard to read. Could you rather split it into 3 components, one for each state?

src/components/atomic-crm/misc/ImportFromJsonDialog.tsx
```diff
@@ -0,0 +1,232 @@
...
+                      </p>
+                      <DownloadErrorFileButton
+                        failedImports={importStatus.failedImports}
+                      />
+                    </>
+                  ) : null}
+                </AlertDescription>
+              </Alert>
+            )}
+            <ImportStats
```
fzaninotto: Question: Why not pass the entire `importStatus` object as prop? It'd simplify types.

src/components/atomic-crm/providers/fakerest/dataProvider.ts
```diff
@@ -108,7 +108,7 @@ async function fetchAndUpdateCompanyData(
   return { ...params, data: newData };
 }
 
-const dataProviderWithCustomMethod: CrmDataProvider = {
+const dataProviderWithCustomMethod: DataProvider & Partial<CrmDataProvider> = {
```
fzaninotto: CrmDataProvider should already contain the DataProvider methods. If not, we need to change that type. 

Also, why the Partial? On the contrary, I think we should definitely conform with the strict CrmDataProvider type here.

src/components/atomic-crm/providers/fakerest/dataProvider.ts
```diff
@@ -158,17 +158,15 @@ const dataProviderWithCustomMethod: CrmDataProvider = {
...
       data: {
         ...data,
         password: "new_password",
       },
     });
 
-    return {
-      ...user.data,
-    };
+    return response;
```
fzaninotto: This is a breaking change. Can you explain why you did it?

src/components/atomic-crm/providers/supabase/dataProvider.ts
```diff
@@ -31,19 +30,25 @@ if (import.meta.env.VITE_SUPABASE_ANON_KEY === undefined) {
...
   apiKey: import.meta.env.VITE_SUPABASE_ANON_KEY,
   supabaseClient: supabase,
   sortOrder: "asc,desc.nullslast" as any,
-});
+  primaryKeys: defaultPrimaryKeys,
+  httpClient: supabaseHttpClient({
+    apiKey: import.meta.env.VITE_SUPABASE_ANON_KEY,
+    supabaseClient: supabase,
+  }),
+  schema: defaultSchema,
```
fzaninotto: Since these are the default, and it used to work before your change, I wonder why you had to change this configuration.

src/components/atomic-crm/providers/supabase/dataProvider.ts
```diff
@@ -31,19 +30,25 @@ if (import.meta.env.VITE_SUPABASE_ANON_KEY === undefined) {
...
 
 const processCompanyLogo = async (params: any) => {
-  let logo = params.data.logo;
+  const logo = params.data.logo;
 
-  if (typeof logo !== "object" || logo === null || !logo.src) {
-    logo = await getCompanyAvatar(params.data);
-  } else if (logo.rawFile instanceof File) {
+  if (logo?.rawFile instanceof File) {
     await uploadToBucket(logo);
```
fzaninotto, line 42: I don't understand how we get the URL of the uploaded logo here. Can you maybe explain that in a comment?

src/components/atomic-crm/providers/supabase/dataProvider.ts
fzaninotto: Why the no-cors here?

src/components/atomic-crm/types.ts
```diff
@@ -60,7 +60,7 @@ export type Company = {
   zipcode: string;
   city: string;
   stateAbbr: string;
-  sales_id: Identifier;
+  sales_id?: Identifier | null;
```
fzaninotto, line 63: Why did you make the sales_id optional? It shouldn't, for Company as for the other entities.

supabase/migrations/20260107140209_contact_avatar.sql
```diff
@@ -0,0 +1,165 @@
+create extension if not exists "http" with schema "extensions";
+
+  create table "public"."favicons_excluded_domains" (
```
fzaninotto: strange indentation

supabase/migrations/20260127140209_imports.sql
```diff
@@ -0,0 +1,165 @@
...
+      );
+
+
+alter table "public"."favicons_excluded_domains" enable row level security;
+
+CREATE UNIQUE INDEX favicons_excluded_domains_pkey ON public.favicons_excluded_domains USING btree (id);
+
+alter table "public"."favicons_excluded_domains" add constraint "favicons_excluded_domains_pkey" PRIMARY KEY using index "favicons_excluded_domains_pkey";
+
+set check_function_bodies = off;
```
fzaninotto, line 17: Why is it necessary? Shouldn't you turn it back on again after the migration?

supabase/migrations/20260127140209_imports.sql
```diff
@@ -0,0 +1,165 @@
...
+ RETURNS text
+ LANGUAGE plpgsql
+AS $function$declare domain_status int8;
+
+begin
+    if exists (select from favicons_excluded_domains as fav where fav.domain = domain_name) then
+        return null;
+    end if;
+
+    return concat(
```
fzaninotto, line 59: Since we execute this server-side, we don't need favicon.show anymore. It was intended to workaround the CORS problems when fetching the ison client-side. We can more simply fetch the `favicon.ico` from the domain. 

Edit: out of scope for this PR.

supabase/migrations/20260107140209_contact_avatar.sql
```diff
@@ -0,0 +1,165 @@
...
+
+    select coalesce(jsonb_array_length(new.email_jsonb), 0) into emails_length;
+
+    if emails_length = 0 then
+        return new;
+    end if;
+
+    for email_entry in select jsonb_array_elements(new.email_jsonb)
+    loop
+        contact_avatar = get_avatar_for_email(email_entry->>'email');
```
fzaninotto: This won't work if there is more than one email I think, as you update a single local variable `contact_avatar`

supabase/migrations/20260107140209_contact_avatar.sql
```diff
@@ -0,0 +1,165 @@
...
+
+grant references on table "public"."favicons_excluded_domains" to "anon";
+
+grant select on table "public"."favicons_excluded_domains" to "anon";
+
+grant trigger on table "public"."favicons_excluded_domains" to "anon";
+
+grant truncate on table "public"."favicons_excluded_domains" to "anon";
+
+grant update on table "public"."favicons_excluded_domains" to "anon";
```
fzaninotto: Why do we grant these rights? It seems a security risk. Please add a comment.

supabase/migrations/20260107140209_contact_avatar.sql
```diff
@@ -0,0 +1,165 @@
...
+grant select on table "public"."favicons_excluded_domains" to "service_role";
+
+grant trigger on table "public"."favicons_excluded_domains" to "service_role";
+
+grant truncate on table "public"."favicons_excluded_domains" to "service_role";
+
+grant update on table "public"."favicons_excluded_domains" to "service_role";
+
+
+  create policy "Enable access for authenticated users only"
```
fzaninotto: indentation problem

supabase/migrations/20260107151133_company_avatar.sql
```diff
@@ -0,0 +1,25 @@
...
+    if company_logo is null then
+        return new;
+    end if;
+
+    new.logo = concat('{"src":"', company_logo, '","title":"Company favicon"}');
+    return new;
+end;$function$
+;
+
+CREATE TRIGGER company_saved BEFORE INSERT OR UPDATE ON public.companies FOR EACH ROW EXECUTE FUNCTION public.handle_company_saved();
```
fzaninotto: I'd move the trigger closer to the function

supabase/migrations/20260108100629_task_due_date_nullable.sql
```diff
@@ -0,0 +1,3 @@
+alter table "public"."tasks" alter column "due_date" drop not null;
```
fzaninotto: Can you merge this with the previous migration?

src/components/atomic-crm/misc/useImportFromJson.ts
```diff
@@ -0,0 +1,687 @@
...
+export type ImportFromJsonStats = {
+  sales: number;
+  companies: number;
+  contacts: number;
+  notes: number;
+  tasks: number;
+};
+
+export type ImportFromJsonFailures = {
+  sales: Array<any>;
```
fzaninotto: can't you be more specific than any?

src/components/atomic-crm/misc/useImportFromJson.ts
```diff
@@ -0,0 +1,687 @@
...
+      sales: {},
+      companies: {},
+      contacts: {},
+      tags: {},
+    };
+
+    const importSale = async (
+      dataToImport: JsonTypes.JsonPrimitive | JsonTypes.JsonStruct | undefined,
+    ) => {
+      if (!isSale(dataToImport)) {
```
fzaninotto: I suggest you move this code down to the try/catch block, and throw an Error if the data isn't a Sale. That way, you won't have to repeat the setState(error)

src/components/atomic-crm/misc/useImportFromJson.ts
```diff
@@ -0,0 +1,687 @@
...
+            name: dataToImport.name.trim(),
+            description: dataToImport.description?.trim(),
+            city: dataToImport.city?.trim(),
+            country: dataToImport.country?.trim(),
+            address: dataToImport.address?.trim(),
+            zipcode: dataToImport.zipcode?.trim(),
+            stateAbbr: dataToImport.state_abbr?.trim(),
+            sales_id: dataToImport.sales_id
+              ? idsMaps.sales[dataToImport.sales_id]
+              : undefined,
```
fzaninotto: No, this should default to the sale actually calling the function.

src/components/atomic-crm/misc/useImportFromJson.ts
```diff
@@ -0,0 +1,687 @@
...
+            contacts: [...old.failedImports.contacts, dataToImport],
+          },
+        }));
+        return;
+      }
+
+      try {
+        let tagsIds: Array<Identifier> = [];
+        if (dataToImport.tags && Array.isArray(dataToImport.tags)) {
+          tagsIds = await Promise.all(
```
fzaninotto, line 291: What if some tags are present but not all?

src/components/atomic-crm/misc/useImportFromJson.ts
```diff
@@ -0,0 +1,687 @@
...
+              : undefined,
+            email_jsonb: Array.isArray(dataToImport.emails)
+              ? dataToImport.emails
+              : undefined,
+            phone_jsonb: Array.isArray(dataToImport.phones)
+              ? dataToImport.phones
+              : undefined,
+            sales_id: dataToImport.sales_id
+              ? idsMaps.sales[dataToImport.sales_id]
+              : undefined,
```
fzaninotto: Same, sales_id should not be undefined

src/components/atomic-crm/misc/useImportFromJson.ts
```diff
@@ -0,0 +1,687 @@
...
+          failedImports: {
+            ...old.failedImports,
+            notes: [...old.failedImports.notes, dataToImport],
+          },
+          error: null,
+        }));
+        return;
+      }
+      try {
+        const defaultSale = await getDefaultSale();
```
fzaninotto: This one should be common to all the import functions, avoid repeating the call.

src/components/atomic-crm/misc/useImportFromJson.ts
```diff
@@ -0,0 +1,687 @@
...
+      contacts: Record<number, Identifier>;
+      tags: Record<string, Identifier>;
+    } = {
+      sales: {},
+      companies: {},
+      contacts: {},
+      tags: {},
+    };
+
+    const importSale = async (
```
fzaninotto, line 139: All the importXXX functions share 70% of their code. I wonder if you could refactor them to use a common core without obscuring the code.

src/components/atomic-crm/misc/useImportFromJson.ts
```diff
@@ -0,0 +1,687 @@
...
+        "$.contacts.*",
+        "$.notes.*",
+        "$.tasks.*",
+      ],
+      keepStack: false,
+    });
+    const stream = file.stream();
+    const reader = stream.pipeThrough(parser).getReader();
+
+    const proccesBatchIfPossible = async (ignoreSize: boolean = false) => {
```
fzaninotto: Can you explain the ignoreSize boolean function? It's not clear by looking at the code.

src/components/atomic-crm/misc/useImportFromJson.ts
```diff
@@ -0,0 +1,687 @@
...
+
+const defaultStats = {
+  sales: 0,
+  companies: 0,
+  contacts: 0,
+  notes: 0,
+  tasks: 0,
+};
+
+export const useImportFromJson = (): [
```
fzaninotto, line 90: I'd add a comment explaining why we do that client-side (the fact that edge functions are limited in memory and run time)

fzaninotto (changes requested):

src/components/atomic-crm/layout/Header.tsx
djhi: The issue is that the `ImportFromJsonButton` does not render once the menu is closed and that would remove the dialog from the DOM as well

src/components/atomic-crm/types.ts
djhi: You asked me to

supabase/migrations/20260127140209_imports.sql
djhi: This is generated by Supabase and no we don't need to turn it back on as it's only applied for the current query (the migration in this case)

src/components/atomic-crm/layout/Header.tsx
fzaninotto: Good point. It might be worth a comment to explain that to future me.

src/components/atomic-crm/misc/useImportFromJson.ts
djhi: This is for failed import which means the records from the json file didn't match the expected format. For what we know, they could indeed be any JSON primitive

src/components/atomic-crm/misc/useImportFromJson.ts
djhi: The array of promises we await validate each tag. Did I miss something?

src/components/atomic-crm/misc/useImportFromJson.ts
djhi: Removed to fallback to currently logged sale

src/components/atomic-crm/providers/fakerest/dataProvider.ts
djhi: That's actually the wrong fix but we have the following issue: the two supabase edge functions that manage sales both return `{ data: sale }` and the Supabase dataProvider returns it untouched. The reason is that `supabase.functions.invoke` returns `{ data: JSONResultFromFunction, error: AnyError}` and `data` will contain the `{ data: sale }` returned by the function. We should either fix the edge functions or the supabase dataprovider. I'll undo this change though

src/components/atomic-crm/providers/supabase/dataProvider.ts
djhi: Left over from when I thought we needed `createMany`. I'll undo these changes

src/components/atomic-crm/providers/fakerest/dataProvider.ts
djhi: Left over from when I thought we needed `createMany`. I'll undo these changes

src/components/atomic-crm/layout/Header.tsx
djhi: Comment added

src/components/atomic-crm/layout/Header.tsx
djhi: Fixed for all menu items

src/components/atomic-crm/misc/ImportFromJsonDialog.tsx
djhi: Added comment, not sure why GH doesn't consider the comment outdated

src/components/atomic-crm/providers/supabase/dataProvider.ts
djhi: If you look at `uploadToBucket`, you'll see that it will download files using their `src` property and upload them on the supabase storage. `useImportFromJson` create attachments by creating objects with the expected `src` property.

supabase/migrations/20260107140209_contact_avatar.sql
djhi: This is generated by supabase migration tool but we have RLS anyway so that shouldn't be a security issue

djhi:
> A few problems I noticed:
> 
> 1. As the JSON import button requires the app to already be initialized with an initial sale, before adding a new sale from the JSON file, check that an existing sale with the same email doesn't exist yet. Otherwise all the contacts and the tasks will be assigned to the wrong sale (and I won't see the tasks that are assigned to me)

This is what I do already: https://github.com/marmelab/atomic-crm/pull/133/files#diff-639b536d8fb9042b244d653a0f37d2774a0cd00158a6e4ad3e44d4556a689b87R146-R154

> 2. It seems the "last seen" field isn't taken into account, so all the contacts are added on January 1st, 1970. This makes taking over on Atomic CRM very difficult.

There's no such field in highrise. Do you want to take the date of the latest note?

> 3. when searching for a company in the company list, I see the following error: 'column companies_summary.stateAbbr does not exist'

Fix incoming

> 4. ~Some attachments weren't fetched~

PR #136, opened 2026-01-15
[Breaking change] Ensure all tables and their fields are named in snake case

djhi:
## Problem

Not all tables nor their fields are named consistently in snake case.

## Solution

Ensure all tables and their fields are named in snake case

## Breaking changes

- table `contactNotes` has been renamed `contact_notes`
- table `dealNotes` has been renamed `deal_notes`
- column `stateAbbr` in table `companies` has been renamed `state_abbr`
- 
You must run the migration to update your database schema:

```sh
make supabase-migrate-database`
```

Also, make sure you update any custom code that referenced the matching resources and fields.

## How To Test

Run the migration on an existing project.
Make sure all CRUD operations still work.

## Additional Checks

### DIFF
--- a/doc/package-lock.json
+++ b/doc/package-lock.json
@@ -11,7 +11,7 @@
         "@astrojs/starlight": "^0.36.0",
         "@astrojs/starlight-tailwind": "^4.0.1",
         "@tailwindcss/vite": "^4.1.14",
-        "astro": "^5.16.6",
+        "astro": "^5.16.10",
         "sharp": "^0.34.2",
         "tailwindcss": "^4.1.14"
       },
@@ -115,7 +115,6 @@
       "resolved": "https://registry.npmjs.org/@astrojs/starlight/-/starlight-0.36.0.tgz",
       "integrity": "sha512-aVJVBfvFuE2avsMDhmRzn6I5GjDhUwIQFlu3qH9a1C0fNsPYDw2asxHQODAD7EfGiKGvvHCJgHb+9jbJ8lCfNQ==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@astrojs/markdown-remark": "^6.3.1",
         "@astrojs/mdx": "^4.2.3",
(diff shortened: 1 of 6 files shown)


src/components/atomic-crm/notes/NoteCreate.tsx
```diff
@@ -84,7 +84,11 @@ const NoteCreateButton = ({
     refetch();
     update(reference, {
       id: (record && record.id) as unknown as Identifier,
-      data: { last_seen: new Date().toISOString(), status: data.status },
+      data: {
+        last_seen:
+          reference === "contacts" ? new Date().toISOString() : undefined,
```
fzaninotto, line 89: why this change?

djhi: This causes errors with deal notes that don't have a `last_seen` field (postgres error)

WiXSL (changes requested):
Some findings

1 -  `supabase/migrations/20251204172855_merge_contacts_function.sql` (line 33): 
merge_contacts SQL function still updates "contactNotes" after the table rename; once 20260115150819_snake_case_renaming.sql runs, this function should fail at runtime.

2 - `supabase/functions/_shared/db.ts` (lines 77-81): 
Kysely schema in edge functions still declares contactNotes, so updateTable("contact_notes") is no longer covered by the typed Database interface and should type-error during function typechecking/deploy.

3 - `src/components/atomic-crm/contacts/TagsListEdit.tsx`  (lines 41-49):
 TagsListEdit now assumes record.tags is always defined. If existing rows have tags NULL (schema has no default), record.tags.includes and spreads will throw

djhi:
> 1 - `supabase/migrations/20251204172855_merge_contacts_function.sql` (line 33): merge_contacts SQL function still updates "contactNotes" after the table rename; once 20260115150819_snake_case_renaming.sql runs, this function should fail at runtime.

Good catch!

> 2 - `supabase/functions/_shared/db.ts` (lines 77-81): Kysely schema in edge functions still declares contactNotes, so updateTable("contact_notes") is no longer covered by the typed Database interface and should type-error during function typechecking/deploy.

Same

> 3 - `src/components/atomic-crm/contacts/TagsListEdit.tsx` (lines 41-49): TagsListEdit now assumes record.tags is always defined. If existing rows have tags NULL (schema has no default), record.tags.includes and spreads will throw

I don't get this one, I haven't touched the schema itself. Aren't you mixing up this PR and #133?

PR #199, opened 2026-03-06
Fix Supabase security advisor warnings about views and functions

slax57:
## Problem

Supabase Advisor raises security warnings after Atomic CRM installation. Some we can fix and some we can't:

- Security Definer View (2)
  - ~~`init_state`~~: expected and [documented](https://marmelab.com/atomic-crm/doc/developers/supabase-configuration/#frequently-asked-questions)
  - `contacts_summary` view
- Function Search Path Mutable (8)
- Extension in Public (1)
- ~~RLS Policy Always True (22)~~: all related tables declare policies requiring the user to be authenticated to use them. We need no futher check (for now) as all sales should be able to view and edit all data in the CRM.
- ~~Leaked Password Protection Disabled (1)~~: depends on developer choice, can be changed in Supabase configuration

## Solutions

- Security Definer View
  - Fix the `contacts_summary` view to enable `security_invoker` (it was done previously in https://github.com/marmelab/atomic-crm/pull/29 but accidentally removed in a later version)
- Function Search Path Mutable
  - Alter or recreate functions to add `SET search_path TO ''`
- Extension in Public
  - Recreate the `pg_net` extension in the `extensions` schema

=> All this is done in a new db migration

## How To Test

- Apply the migration, either locally or remotely, and confirm the amount of warnings decreases from 34 to 24
- Test the app to ensure functions are still working

## Additional Checks


supabase/migrations/20260306152531_fix_security_warnings.sql
```diff
@@ -0,0 +1,355 @@
...
+declare email_hash text;
+declare gravatar_url text;
+declare gravatar_status int8;
+declare email_domain text;
+declare favicon_url text;
+declare domain_status int8;
+
+begin
+    -- Try to fetch a gravatar image
+    email_hash = encode(digest(email, 'sha256'), 'hex');
```
WiXSL:
```suggestion
    email_hash = encode(extensions.digest(email, 'sha256'), 'hex');
```

`digest()` is no longer resolved here after `SET search_path TO ''`.
I reproduced it locally: this function returns ERROR

WiXSL (changes requested):

supabase/migrations/20260309112831_fix_security_warnings.sql
```diff
@@ -0,0 +1,355 @@
...
+    co.last_seen,
+    co.has_newsletter,
+    co.status,
+    co.tags,
+    co.company_id,
+    co.sales_id,
+    co.linkedin_url,
+    c.name as company_name,
+    count(distinct t.id) as nb_tasks
+from
```
fzaninotto, line 18:
```suggestion
select 
    co.*,
    jsonb_path_query_array(co.email_jsonb, '$[*].email')::text as email_fts,
    jsonb_path_query_array(co.phone_jsonb, '$[*].number')::text as phone_fts,
    c.name as company_name,
    count(distinct t.id) as nb_tasks
from
```

is easier to read

supabase/migrations/20260306152531_fix_security_warnings.sql
```diff
@@ -0,0 +1,355 @@
...
+    companies c on co.company_id = c.id
+group by
+    co.id, c.name;
+
+-- Ensure all functions have search_path set to empty
+
+-- Functions that already use qualified names can be simply altered
+
+ALTER FUNCTION public.get_note_attachments_function_url()
+SET search_path TO '';
```
fzaninotto: My understanding is that there must be at least `public` in the search path, as that's the schema where we defined the function. 

Supabase advises this resolution:

> Choose a minimal, explicit path—typically just `pg_catalog`, `public` or even just `public` if everything is schema-qualified.

supabase/migrations/20260306152531_fix_security_warnings.sql
```diff
@@ -0,0 +1,355 @@
...
+
+ALTER FUNCTION public.get_user_id_by_email(text)
+SET search_path TO '';
+
+ALTER FUNCTION public.handle_contact_saved()
+SET search_path TO '';
+
+-- get_avatar_for_email: needs to be recreated with qualified names
```
fzaninotto: You should make the comment a bit more explicit about what wasn't qualified in the initial function (I had to diff by hand to find out): `digest`, `http_get`, `get_domain_favicon`

Same for the other recreated functions

supabase/migrations/20260306152531_fix_security_warnings.sql
```diff
@@ -0,0 +1,355 @@
+-- Recreate the contacts_summary view with security_invoker enabled
```
fzaninotto: Would you mind adding an initial comment explainning the goal of the entire migration?

fzaninotto (changes requested):
This type of PR is super hard to review, as we have to diff the previous and new SQL by hand. We need to fin-d a way to make it easier.

supabase/migrations/20260306152531_fix_security_warnings.sql
slax57: wow nice catch!

supabase/migrations/20260306152531_fix_security_warnings.sql
slax57: Not exactly sure about what you mean here, but `search_path` affects the way postgest *resolves* the names in the function implementation, but does not affect where the function itself is declared.

I have confirmed by testing that setting the search_path to an empty string works (because all functions are now using fully qualified names, and the functions that didn't require recreating either already used fully qualified names, or didn't have to resolve names at all).

slax57: UPDATE: following discussion:
- we keep `SET search_path TO ''` for functions that had it
- we use `SET search_path TO 'public'` for altered functions (to simplify the code and comply with Supabase best practices)

WiXSL (approved):

PR #219, opened 2026-03-18
Fix Inbound Email: case-insensitive contact email search and improved company creation

slax57:
### 1. Contact email search is case-sensitive

1. I have a contact named John Doe with email john.doe@example.com
2. I write to [John.Doe@example.com](mailto:John.Doe@example.com "‌") (notice the capitalization) and CC Atomic cRM
3. Atomic CRM will create a new contact instead of attaching the message to the old contact

This is a real life problem, as some contacts send emails with capitl letters in it and we just reply to these messages.

### 2. Companies are created with the full domain name

When we forward an email to Atomic CRM and the contact doesn’t exist, Atomic CRM creates it and also creates a company for it based on the domain.

But it uses the full domain name (e.g. “john@example.com” is attached to the “[example.com](http://example.com "‌")” company instead of “Example”)

a PR was started a long time ago to fix this: [https://github.com/marmelab/atomic-crm/pull/62](https://github.com/marmelab/atomic-crm/pull/62 "")

## Solutions

1. Store emails and websites in lowercase in DB (using triggers)
2. Update existing values to be lowercase
3. Use lowercased emails and websites when searching for contacts and companies
4. Strip the TLD from the domain to compute the company name (using [tldts](https://www.npmjs.com/package/tldts))
5. Store the company domain as website instead (and search by website)

## How To Test

- Migrations
  - add sales, contacts and companies with various casing for emails and websites
  - run `make supabase-migrate-database`
  - Confirm all is now lowercase
- Triggers
  - add sales, contacts and companies with various casing for emails and websites 
  - Confirm they are stored lowercased
- Webhook
  - Follow the instructions [here](https://github.com/marmelab/atomic-crm/blob/main/supabase/functions/postmark/index.ts#L173) to test the webhook
  - Try using various casing for contact, sales and inbound emails
  - Try both the Cc: and Fwd: examples
  - Confirm new contacts and companies are only created when needed
  - Confirm companies are created with the correct name and website
  - Confirm companies with no website can still be matched by name (derived from the domain)

## Additional Checks


supabase/functions/postmark/addNoteToContact.ts
```diff
@@ -78,7 +78,7 @@ export const getOrCreateContactFromEmailInfo = async ({
     .insert({
       first_name: firstName,
       last_name: lastName,
-      email_jsonb: [{ email, type: "Work" }],
+      email_jsonb: [{ email: email.toLowerCase(), type: "Work" }],
```
WiXSL: We should also normalize the domain before company lookup/creation; otherwise, emails like GMAIL.COM / ACME.COM can still create bogus or duplicate companies

WiXSL (changes requested):

supabase/functions/postmark/compareIgnoringCase.ts
```diff
@@ -0,0 +1,4 @@
+export const compareIgnoringCase = (
```
fzaninotto: Please do not introduce new files unless the feature is shared. Each new file has a cost for updates and the registry.

fzaninotto (changes requested):

supabase/functions/postmark/compareIgnoringCase.ts
slax57: It IS shared (used in 2 different files)

supabase/functions/postmark/addNoteToContact.ts
```diff
@@ -11,7 +12,7 @@ export const getOrCreateCompanyFromDomain = async (
     await supabaseAdmin
       .from("companies")
       .select("*")
-      .eq("name", domain)
+      .ilike("name", escapeSqlWildcards(domain))
```
fzaninotto: The problem is that users can update the company name, and we also want to have a default name that doesn't look like a domain name (e.g., "Acme" instead of "acme.com"). 

So instead of comparing a domain with a name, we should compare the domain with the company URL

supabase/functions/postmark/addNoteToContact.ts
```diff
@@ -23,14 +24,16 @@ export const getOrCreateCompanyFromDomain = async (
...
+    MAIL_PROVIDERS.some((provider) => compareIgnoringCase(provider, domain))
+  ) {
     // We don't want to create companies for generic mail providers, as they are not really companies and it would pollute the database with useless entries.
     return null;
   }
 
   const { data: newCompanies, error: createCompanyError } = await supabaseAdmin
     .from("companies")
-    .insert({ name: domain, sales_id: salesId })
+    .insert({ name: domain.toLowerCase(), sales_id: salesId })
```
fzaninotto: The name should be humanized, and we should set the company URL with the real domain instead

supabase/functions/postmark/addNoteToContact.ts
```diff
@@ -56,9 +59,9 @@ export const getOrCreateContactFromEmailInfo = async ({
   // Check if the contact already exists
   const { data: existingContact, error: fetchContactError } =
     await supabaseAdmin
-      .from("contacts")
+      .from("contacts_summary")
       .select("*")
-      .contains("email_jsonb", JSON.stringify([{ email }]))
+      .ilike("email_fts", `%"${escapeSqlWildcards(email)}"%`)
```
fzaninotto: why the `%`? We want to compare exact matches but in a case insensitive way

Also, a better way IMO would be to force the view to output lowercase emails, and to lowercase the email we search so we can keep an exact comparison (faster)

supabase/functions/postmark/index.ts
fzaninotto, line 57: you could lowercase here to make the search simpler

supabase/functions/postmark/index.ts
```diff
@@ -61,15 +62,20 @@ Deno.serve(async (req) => {
   if (
     INBOUND_EMAIL &&
     ToFull.length === 1 &&
-    ToFull[0].Email === INBOUND_EMAIL &&
-    salesEmails.includes(FromFull.Email)
+    compareIgnoringCase(ToFull[0].Email, INBOUND_EMAIL) &&
```
fzaninotto: I think you'd better lowercase the FromFull.Email and ToFull.email instead.

fzaninotto (changes requested):

supabase/functions/postmark/addNoteToContact.ts
slax57: Improving the company name is already planned in another card of our internal BL. I suggest we fix both issues in a new PR, when addressing this other card.

supabase/functions/postmark/addNoteToContact.ts
slax57: same (other PR)

supabase/functions/postmark/addNoteToContact.ts
slax57: `email_fts` is an array of email addresses (stringified), hence the need for `%`. Comparision is still exact thanks to the `"` tokens I added.

Keep in mind this `ilike` search is only performed when we process an inbound email, whereas the `contacts_summary` view is used much more often. IMO it's not that obvious that it would improve performances to lowercase the emails in the view. Wdyt?

supabase/functions/postmark/index.ts
slax57: I thought it'd make the code more robust to support casing differences in the sales email and inbound email addresses too. But if you think that's going too far, then yes I can simply lowercase before searching.

supabase/functions/postmark/index.ts
slax57: same (about robustness)

supabase/schemas/04_triggers.sql
```diff
@@ -38,6 +38,21 @@ create or replace trigger contact_saved
...
 
+-- Lowercase company website before insert or update
+create or replace trigger lowercase_company_website
+    before insert or update on public.companies
+    for each row execute function public.lowercase_website();
+
+-- Lowercase contact emails before insert or update
+create or replace trigger lowercase_contact_emails
+    before insert or update on public.contacts
+    for each row execute function public.lowercase_email_jsonb();
```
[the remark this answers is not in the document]
WiXSL: Copilot is mostly right here. 
The key issue is trigger ordering: this change assumes `lowercase_contact_emails` will run before `contact_saved`, but PostgreSQL executes same-event BEFORE triggers in NAME order, not creation order.

WiXSL (changes requested):

supabase/schemas/04_triggers.sql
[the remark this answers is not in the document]
slax57: Right! Copilot incorrectly suggested me to fix this by changing the trigger creation order, but after checking in the Postgres docs you are correct, only the alphabetical order matters.
Should be fixed by https://github.com/marmelab/atomic-crm/pull/219/commits/439a7080e3f825f440a1b7c985643da8053d2216

WiXSL (approved):

supabase/functions/postmark/addNoteToContact.ts
```diff
@@ -2,16 +2,28 @@ import { supabaseAdmin } from "../_shared/supabaseAdmin.ts";
...
+    return null;
+  }
+
   // Check if the company already exists
   const { data: existingCompany, error: fetchCompanyError } =
     await supabaseAdmin
       .from("companies")
       .select("*")
-      .eq("name", domain)
+      .or(`website.eq.${website},name.eq.${companyName},name.eq.${domain}`)
```
fzaninotto, line 26: *Question*: how can name equal domain ?

supabase/functions/postmark/extractMailContactData.ts
```diff
@@ -1,5 +1,24 @@
...
+export function extractCompanyName(domain: string) {
+  if (!domain) return "";
+  let name = domain.split(".").at(-2) || domain;
+
+  const parseResult = parse(domain);
+  if (parseResult.domainWithoutSuffix) {
+    name = parseResult.domainWithoutSuffix;
+  }
+
+  const humanizedName = name
```
fzaninotto, line 14: `inflection` is a dependency of ra-core, so you could use `inflecion.humanize` instead.

supabase/functions/postmark/extractMailContactData.ts
```diff
@@ -1,5 +1,24 @@
+import { parse } from "npm:tldts";
```
fzaninotto: since you added the dependency in `deno.json`, it's not necessary to resolve it here, so you can use `tldts` instead of `npm:tldts`, or you can remove the entry in `deno.json`

fzaninotto (changes requested):
Nearing completion!

supabase/functions/postmark/extractMailContactData.ts
slax57: Yes but not a dep of Edge functions runtime. I did not feel like adding it just for this simple task.

supabase/functions/postmark/addNoteToContact.ts
slax57: for older companies created by the previous implementation of the Inbound Email

supabase/functions/postmark/addNoteToContact.ts
```diff
@@ -88,7 +104,7 @@ export const getOrCreateContactFromEmailInfo = async ({
     .select();
   if (createContactError || !newContacts[0]) {
     throw new Error(
-      `Could not create contact in database, email: ${email}, error: ${createContactError.message}`,
+      `Could not create contact in database, email: ${email}, error: ${createContactError?.message}`,
```
ThieryMichel: in this block createContactError is sure to exists. And ts infer it correctly.
```suggestion
      `Could not create contact in database, email: ${email}, error: ${createContactError.message}`,
```

ThieryMichel (changes requested):

supabase/functions/postmark/addNoteToContact.ts
slax57: Actually no, I believe this check **is** necessary because the error can also be thrown if `createContactError` is undefined but `newContacts[0]` is undefined too. So I force pushed to discard the last commit.

ThieryMichel (approved):

fzaninotto (approved):

PR #232, opened 2026-03-25
Migrate to Vite 8

WiXSL:
Waiting for rolldown to be more stable

The migration is working technically: the project installs, typechecks, and builds successfully.

That said, there are still ecosystem compatibility gaps around Vite 8, so this PR is worth reviewing carefully before merge.

## What was updated

- upgraded `vite` to v8
- upgraded `@vitejs/plugin-react`
- upgraded `vitest` and related browser packages
- upgraded `storybook` / `@storybook/react-vite`
- adjusted the Vite config for the new build pipeline
- updated `.storybook/main.ts` for ESM execution

## Validation

I verified that these commands pass:

- `npm ci --ignore-scripts`
- `npm run typecheck`
- `npm run build`
- `npm run build:demo`
- `npm run build-storybook`

I also verified the PWA flow still works:

- production build still generates:
  - `dist/registerSW.js`
  - `dist/sw.js`
  - `dist/workbox-*.js`
- `dist/index.html` still includes:
  - the manifest link
  - `registerSW.js`
- local Chromium check:
  - service worker registers successfully
  - `sw.js` and `workbox-*.js` are requested
- offline check:
  - after one online load, the app shell still loads offline under service worker control

### `vite-plugin-pwa`
`vite-plugin-pwa@1.2.0` still declares Vite support only up to `^7`.

This PR uses an `overrides` entry so npm resolves `vite-plugin-pwa` against the project's Vite 8 version.

### Storybook transitive dependency
`@storybook/react-vite` still pulls in `@joshwooding/vite-plugin-react-docgen-typescript` through a dependency range that does not resolve to the Vite 8-compatible version by default.

This PR uses an `overrides` entry in `package.json` so that the resolved version is compatible with Vite 8.

### Explicit Workbox packages
This PR keeps explicit `workbox-build` and `workbox-window` dependencies because `vite-plugin-pwa` requires them at build time in this setup.


vendor/vite-plugin-pwa-1.2.0.tgz
ThieryMichel, line 1: Why do we commit this file ?
vite-plugin-pwa 1.2.0 is published on npm.

package.json
```diff
@@ -124,13 +124,18 @@
...
+    "storybook": "^10.3.3",
     "typescript": "~5.8.3",
     "typescript-eslint": "^8.35.1",
-    "vite": "^7.3.0",
-    "vite-plugin-pwa": "^1.2.0",
-    "vite-plugin-simple-html": "^1.0.1",
-    "vitest": "^4.0.18",
-    "vitest-browser-react": "^2.0.5"
+    "vite": "^8.0.2",
+    "vite-plugin-pwa": "file:vendor/vite-plugin-pwa-1.2.0.tgz",
```
ThieryMichel: You can install 1.2.0 directly:

```suggestion
    "vite-plugin-pwa": "^1.2.0",
```

ThieryMichel (changes requested):

vendor/vite-plugin-pwa-1.2.0.tgz
WiXSL: The npm package `vite-plugin-pwa@1.2.0` still declares Vite support only up to `^7`, so installing it with `vite@8` fails with `ERESOLVE`.

`vendor/vite-plugin-pwa-1.2.0.tgz` is a patch version used as a workarround.

But I found a mutch better soluton for this: https://github.com/vite-pwa/vite-plugin-pwa/issues/918

vite.config.ts
```diff
@@ -24,6 +24,7 @@ export default defineConfig({
       },
     }),
     VitePWA({
+      injectRegister: "script",
```
ThieryMichel: We do  not need this, it works without it.

ThieryMichel (dismissed):

package-lock.json
```diff
@@ -16438,267 +14575,261 @@
...
+        "node": ">=0.10.0"
       }
     },
-    "node_modules/set-function-name": {
-      "version": "2.0.2",
-      "resolved": "https://registry.npmjs.org/set-function-name/-/set-function-name-2.0.2.tgz",
-      "integrity": "sha512-7PGFlmtwsEADb0WYyvCMa1t+yke6daIG4Wirafur5kcf+MhUnPms1UeR0CKQdTZD81yESwMHbtn+TR+dMviakQ==",
-      "dev": true,
+    "node_modules/rolldown": {
+      "version": "1.0.0-rc.12",
```
fzaninotto, line 14787: I'm a bit uncomfortable relying on a core dep that isn't stable yet

fzaninotto:
rolldown v1 has been released. Can you update the PR? Also, needs rebase.

PR #246, opened 2026-04-01
Persist user preferences (theme and locale) in database

slax57:
## Problem

User preferences (theme and locale) are stored locally and lost when switching devices or browsers, or after logging out.

## Solution

Persist user preferences in a `preferences` JSONB column on the `sales` table, synced on change and pre-fetched at login.

## How To Test

- Change theme and locale, log out and back in — preferences should be restored
- Log in from a different browser — preferences should be restored (Supabase only)

## Additional Checks


src/components/atomic-crm/settings/SettingsPageMobile.tsx
```diff
@@ -405,6 +412,12 @@ const LanguageRow = () => {
 const ThemeRow = () => {
   const translate = useTranslate();
   const { theme, setTheme } = useTheme();
+  const dataProvider = useDataProvider<CrmDataProvider>();
+
+  const handleSetTheme = (value: Theme) => {
+    setTheme(value);
+    dataProvider.updatePreferences({ theme: value });
+  };
```
ThieryMichel, line 424: Now that it's async and depends on the server, I think we should add a notification on success and on error.

src/components/atomic-crm/settings/SettingsPageMobile.tsx
```diff
@@ -371,6 +372,12 @@ const LanguageRow = () => {
   const translate = useTranslate();
   const locales = useLocales();
   const [locale, setLocale] = useLocaleState();
+  const dataProvider = useDataProvider<CrmDataProvider>();
+
+  const handleSetLocale = (value: string) => {
+    setLocale(value);
+    dataProvider.updatePreferences({ locale: value });
+  };
```
ThieryMichel, line 384: We should add notification and error handling.

ThieryMichel (changes requested):

src/components/atomic-crm/settings/SettingsPageMobile.tsx
slax57: To me this persistence should be non-blocking. However I agree the user should be notified in case of error. Errors were already logged to the console, but I added a proper error notification for each button now.

src/components/atomic-crm/settings/SettingsPageMobile.tsx
slax57: same

src/components/atomic-crm/settings/SettingsPageMobile.tsx
ThieryMichel: I still think that now that it triggers a request it should at least display a loader on the input. (It materialize that some works is happening and would avoid edge case where the user do a second change before the first one complete)

ThieryMichel (changes requested):

src/components/atomic-crm/settings/SettingsPageMobile.tsx
pysnooLab: Fixed by chaining the writes in a single helper all five controls go through, with a regression test that fails without it. No loader though: the change is applied instantly and the write is a background sync, so a spinner would suggest the choice isn't taken yet when it already is.

Fixed error toast, including the zero-row case PostgREST reports as a 204, and a failed write rolls the value back visibly. Success toast left out, it's noise for an immediately visible change.

pysnooLab:
Rebased on main. 
The write mechanism is now centralised in one helper while each control still calls it explicitly
The language revert, a privilege escalation through the new self-update policy, and demo mode are fixed.

erwanMarmelab:
need rebase

PR #274, opened 2026-04-16
Add Task list MCP App

fzaninotto:
## Problem

The MCP server only replies with text, which makes interactions slow. 

## Solution

Reply to queries about task lists with a special task list component, leveraging the MCP Apps standard. 

image


fzaninotto:
I had to merge this PR to test the feature in a real agent. I discovered that using Vue via an external CDN isn't possible due to CSP, so I had to rewrite the MCP app to vanilla JS in bcebded9a16970a46de1f41d56cb5034be7e95fd.

PR #292, opened 2026-06-01
fix(supabase): Add grants for views to migrations

jonathanarnault:
## Problem

`init_state` is not available when supabase is configured to hide public views without grant.

## Solution

Add migrations for these grants

### DIFF
--- a/supabase/migrations/20260601120000_grant_init_state_to_api_roles.sql
+++ b/supabase/migrations/20260601120000_grant_init_state_to_api_roles.sql
@@ -1,0 +1,33 @@
+-- The CRM's public views (activity_log, companies_summary, contacts_summary
+-- and init_state) are queried through the Supabase Data API by the anon and
+-- authenticated roles. `init_state` in particular is read by the anon role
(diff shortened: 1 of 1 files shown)


erwanMarmelab (approved):

fzaninotto:
Just flagging a least-privilege point for a possible follow-up (this doesn't block the fix — the migration is safe as-is):

Only `init_state` actually needs to be reachable by `anon`. The other three views (`activity_log`, `companies_summary`, `contacts_summary`) are queried after login, so they only need `authenticated` + `service_role`.

The reason granting them to `anon` is harmless today is that all three are declared with `security_invoker = on`, while their underlying tables have RLS policies scoped `to authenticated` only (no `anon` policy). So an `anon` request to those views passes the table-level grant but returns zero rows. `init_state` is the deliberate exception: `security_invoker = off` (bypasses RLS) and exposes only an `is_initialized` count.

So no data leaks — but the `anon` grants on the three summary views are dead privileges. If we want to tighten things, we could drop them from both `06_grants.sql` (the source of truth, which already has them at lines 110-124) and a follow-up migration, keeping only:

​```sql
grant all on table public.init_state to anon;
​```

PR #294, opened 2026-06-02
feat(skill): delete-resource

erwanMarmelab:
## Problem

The agent has to read a lot of code to delete existing features which is quite costly

## Solution

Provide skills that help agents delete existing features


.claude/skills/delete-initial-resource/SKILL.md
jonathanarnault, line 1: This skill is veerrrrry large, can you make it more concise ?

Also, It refers to node 18, but the package.json references node 22 in its types, can you fix that ?

jonathanarnault (changes requested):

.claude/skills/delete-initial-resource/SKILL.md
pysnooLab: Maybe you can split by Entity like : 
delete-initial-ressource/
  SKILL.md          ← Main entry
  core.md           ← common rules
  tasks.md      ← specific case for tasks
  companies.md      ← specific case for comanies

And in SKILL.md
# Skill delete-initial-ressource

## Proceed
1. Always read `core.md` first
2. Identify cases : 
   - Tasks → read `tasks.md`
   - Companies → read `companies.md`

jonathanarnault (approved):

fzaninotto (changes requested):
since core.md must always be read, I don't see the point of extracting it to a separate file. Besides, the core SKILL file repeats the same information 2 or 3 times, it's not necessary

.claude/skills/delete-initial-resource/delete-initial-resource.ts
```diff
@@ -0,0 +1,458 @@
...
+const initialResource = [
+  "contacts",
+  "companies",
+  "deals",
+  "tags",
+  "tasks",
+] as const;
+const dependentFiles: ResourceFiles = {
+  companies: [
+    // `companies` is a SPINE/LINK resource like contacts: it owns the companies
```
pysnooLab: Do you need all these comments before each entities ? Same information is already in md files. If there is a change, we will have to modify comments at multiple places.

.claude/skills/delete-initial-resource/SKILL.md
```diff
@@ -0,0 +1,64 @@
...
+# delete-initial-resource
+
+Removes one or more of the five initial Atomic CRM resources (`contacts`, `companies`, `deals`, `tags`, `tasks`) and every reference to them. **Irreversible — confirm the target(s) with the user first; rely on git to recover.**
+
+## Steps
+
+1. **Confirm the target(s)**, then read each one's file before editing:
+   - `contacts` → [`contacts.md`](contacts.md) — spine; **confirm cascade scope first**
+   - `companies` → [`companies.md`](companies.md) — spine/link; **confirm cascade scope first**
+   - `deals` → [`deals.md`](deals.md) · `tags` → [`tags.md`](tags.md) · `tasks` → [`tasks.md`](tasks.md)
```
fzaninotto: why are tags and tasks in the same line?

package.json
```diff
@@ -105,7 +105,7 @@
     "@types/jsonexport": "^3.0.5",
     "@types/lodash": "^4.17.20",
     "@types/ms": "^2.1.0",
-    "@types/node": "^22.16.5",
+    "@types/node": "^22.19.19",
```
fzaninotto, line 108: unrelated change. Why did you need this?

.claude/skills/delete-initial-resource/companies.md
```diff
@@ -0,0 +1,30 @@
+# Deleting `companies`
+
+A **spine/link** resource: it owns the `companies` table *and* is a `company_id` link on **both** `contacts` and `deals`, its `companies_summary` view aggregates `nb_contacts`/`nb_deals`, and it adds a **denormalized `company_name`** to the `Contact` type. So it fans out far wider than its folder (≈50 frontend files + backend).
+
+## Confirm cascade scope first
+Ask whether to:
+- **remove `company_id` fully** (cleanest — drops the column from the `Contact`/`Deal` types, the CSV import/export, `mergeContacts`, the fakerest generators, and i18n), or
+- **leave it as an orphaned shell** (keep the column typed `Identifier`, default `null`/unset, but delete all company UI — compiles, but leaves dead FKs pointing at a dropped table).
```
fzaninotto: in which case would this be desireable?

fzaninotto (dismissed):
Solid skill, great guidance.

package.json
erwanMarmelab: Because we need a node version > 22.18 to be able to run a typescript script without `--experimental-strip-types`
https://nodejs.org/en/blog/release/v22.18.0

.claude/skills/delete-initial-resource/delete-initial-resource.ts
erwanMarmelab: removed

PR #310, opened 2026-06-19
Feat: add claude LSP for TypeScript

jonathanarnault:
## Problem

Code is hard to navigate for agents which relies a lot on grep for that

## Solution

Add TypeScript Language Server to help agents navigate the code

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -5,7 +5,7 @@
     "playwright@claude-plugins-official": false,
     "security-guidance@claude-plugins-official": false,
     "supabase@claude-plugins-official": false,
-    "typescript-lsp@claude-plugins-official": false,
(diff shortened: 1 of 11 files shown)


pysnooLab:
Can you add your metrics comparison with or without LSP in description of PR please

.claude/settings.json
```diff
@@ -14,6 +14,9 @@
     "ralph-loop@claude-plugins-official": false
```
erwanMarmelab, line 14: Shouldn't we remove the others ?

jonathanarnault: I think we need that so if the dev has them, we don't want them to be used

erwanMarmelab (approved):

PR #337, opened 2026-07-29
feat: Create deal from contact

erwanMarmelab:
## Feature Idea

On a contact's detail page, it would be nice to have a button to create a deal linked to that person.

## Solution

Add a `Create deal` button an the contact's show page

## Additional Checks

- [x] Tested with **Mobile** resolution -> _not present on mobile_


fzaninotto:
The size and position of this button currently gives it too much importance relative to the other actions IMHO. 

image

I'd prefer that you add a "deals" section in the sidebar listing the current deals with this contact, and move the "add deal" button in this section (and make it a bit smaller like the other "add tasks" buttons). As deals aren't ported to mobile, this section should be hidden on Mobile.

src/components/atomic-crm/contacts/ContactDealsList.tsx
```diff
@@ -0,0 +1,83 @@
...
+ * postgrest provider always builds `getManyReference` targets as `<target>=eq.<id>`,
+ * which an array column cannot match.
+ */
+export const ContactDealsList = () => {
+  const contact = useRecordContext<Contact>();
+
+  if (!contact) return null;
+
+  return (
+    <ListBase
```
fzaninotto, line 20: Using the `error` and `loading` prop together with a `RecordsIterator` (https://marmelab.com/ra-core/recordsiterator/) as child you wouldn't even need a separate DealsIterator.

src/components/atomic-crm/contacts/ContactDealsList.tsx
```diff
@@ -0,0 +1,83 @@
...
+                options={{
+                  notation: "compact",
+                  style: "currency",
+                  currency,
+                  currencyDisplay: "narrowSymbol",
+                  minimumSignificantDigits: 3,
+                }}
+              />
+            </span>
+          </Link>
```
fzaninotto: We already have something almost identical in `ContactDealList`. Why don't you refactor it into a common `DealList` placed under the `deal` directory instead?

fzaninotto (changes requested):

fzaninotto:
You've somehow changed the font size of the items in the company deal list. Beware that the dealiterator probably has two display modes: "normal" for the company show page, and "small" for the contact sidebar.

fzaninotto:
You probably also need to update `.claude/skills/delete-initial-resource/deals.md` because removing the deals will require some cleanup in the contactt show page

PR #338, opened 2026-08-04
Type check with the TypeScript 7 native compiler

pysnooLab:
## Problem

TypeScript 7 (GA on 2026-07-08) is the Go compiler and is around 10x faster, but it cannot simply replace `typescript@5.8`: it no longer ships the JS API that `typescript-eslint` needs. Upstream closed the TypeScript 7 support request as `not planned` until that API lands in 7.1, so a plain bump breaks ESLint on every `.ts`/`.tsx`.

The codebase was not TS 7 ready either: `baseUrl` has been removed from the compiler options, and `getList`/`getOne` were redeclared **non generic** in the Supabase data provider while several call sites pass a type argument (`getOne<Sale>`). TS 5.8 accepted that, TS 7 rejects it with TS2558.

## Solution

The side by side setup documented in the TypeScript 7.0 announcement:

- `tsc` comes from `typescript@7` through the `@typescript/native` alias
- the `typescript` entry resolves to `@typescript/typescript6`, so ESLint keeps a supported peer
- `typescript-eslint` floor raised to `^8.58.0`, the first release whose peer range widens to `typescript <6.1.0`. Below that, `npm ci` refuses the resolution outright.

The two binaries do not collide (`tsc` for 7, `tsc6` for 6), so **no script and no CI job changed**. Removing the alias is all it takes once `typescript-eslint` supports 7.1.

Plus the compatibility fixes, all of which stay valid on TS 5.8:

- `baseUrl` dropped from both tsconfigs (`paths` now resolve relative to the tsconfig file, and `tsconfig-paths` 4.2 used by the Storybook docgen reports the same `absoluteBaseUrl` without it)
- `getList`/`getOne` made generic in the Supabase and FakeRest providers, which fixes the call sites at the root instead of stripping their type arguments
- `scrollMargin` added to the `IntersectionObserver` stub in the notes test, now required by the lib.dom shipped with TS 7

Measured here, 5 runs each: type checking drops from a 4.66s median to 1.10s (4.2x). The build is unchanged, `vite build` does not go through `tsc`.

Two known limits, both dormant:

- `tsconfig.node.json` (the `vite.config.ts` project) fails under TS 7 because the program pulls in `node_modules/@swc/html/index.ts`, a raw `.ts` shipped by a third party, incompatible with `verbatimModuleSyntax`. No script or CI job targets that project today.
- The editor needs the `TypeScriptTeam.native-preview` extension (VS Code >= 1.126) to get the native language server, it is not built into VS Code yet as of 1.131.

## How to test

```sh
npm ci
npx tsc --version   # Version 7.0.2
npm run typecheck   # clean
npm run lint        # clean, ESLint still runs every rule
npm run test        # 405 passed on app + claude, 108 on functions
```


tsconfig.json
erwanMarmelab, line 5: `tsconfig.node.json` no longer compiles under TS 7, so `vite.config.ts` is untypecheckable.

package.json
erwanMarmelab, line 11: `tsc` type-checks nothing (`files: []`). We need to find another way (like `npm run typecheck`. But without the `npm` to let users use their own package manager)

package-lock.json
erwanMarmelab, line 1: I still can find `typescript6` in the `package-lock`

erwanMarmelab (changes requested):

PR #340, opened 2026-08-04
Add project name for telemetry

pysnooLab:
Add project name for telemetry

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -16,7 +16,8 @@
   "skillListingMaxDescChars": 512,
   "env": {
     "ENABLE_LSP_TOOL": "1",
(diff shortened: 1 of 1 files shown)


.claude/settings.json
erwanMarmelab, line 1: Without `CLAUDE_CODE_ENABLE_TELEMETRY=1`, it won't export anything, right?

pysnooLab: Yes, but CLAUDE_CODE_ENABLE_TELEMETRY=1 is already set org-wide through the server-managed settings, see ~/.claude/remote-settings.json.

pysnooLab:
Only need to be local

PR #341, opened 2026-08-04
Move the harness e2e suite into a hook, and cover the untested guards

pysnooLab:
## Problem

Update of harness. Six independent defects, several of them silent:

- `harness.config.json` declared a `cwd: "repo"` e2e step, so in full mode the per-ticket SubagentStop chain ran Playwright in the base-branch checkout: it tested code the ticket had not touched, while a ticket can legitimately be mid-feature. `orchestrator.md` already documented the opposite, so config and docs contradicted each other. `playwright.config.ts` has no `webServer`, so that step could only ever pass by accident.
- `bash-guard` required a `--headless` flag that no Playwright subcommand accepts (1.60 defines it only on the hidden `run-test-mcp-server`; `test` has `--headed` and is headless by default). It therefore blocked every legitimate run, including any command whose path merely contains `test`, while letting `playwright open` through.
- Hook tests leaked session dirs into the real `/tmp`. A green run left four new directories, one under the actual repo path, indistinguishable from a live session's state. ~226 strays had accumulated.
- `ensure-playwright-mcp` only checked that `@playwright/mcp/cli.js` existed, so it reported "present" while the Chromium it needs was absent. In that state the MCP falls back to a `chrome` channel that provisioning never installs, every `browser_navigate` dies, and the reviewer is told to treat a missing binary as a sandbox limitation and never a `REJECTED`. The whole feature-smoke leg was dead in silence.
- `completion-invariant` built the verdict-flag path by hand from `ctx.ticketsDir`, resolving to `<session_dir>/tickets/reviews/`. `TICKETS_DIR` is `<session_dir>`, so the flags live at `<session_dir>/reviews/` and that directory never existed: the APPROVED filter matched nothing and **the orphan invariant never fired**.
- Five hooks had no tests at all, and all five were guards. The bug above is what an untested guard looks like.

## Solution

- Drop the e2e step from the per-ticket chain; add `e2e` to `validation.extraForbidden` so the guard keeps blocking manual runs.
- New `e2e-on-feature-review.mjs` (SubagentStop) is now the only place the suite is launched. It fires on the `MODE: feature-review` stop and only when that review APPROVED, keyed on the reviewer's own flag rather than a verdict parse (`last_assistant_message` is absent in this runtime). Results land in `<session_dir>/e2e-result.json`, stamped with the session-branch sha so a verdict from an earlier request cannot be read as current. `bash-guard` now refuses e2e for **every** caller, orchestrator included, still config-driven.
- Prescribe what a red suite means: same fix recipe as a `BLOCKED:` review, with a shared 2-round bound, and `completion-invariant` rejects the orchestrator's stop once if it stops without acting on it.
- Fix the Playwright rule to block what actually opens a window (`--headed` / `--ui` / `--debug`, `open` / `codegen`).
- Redirect the hooks' tmp root once for the whole `claude` vitest project; rename the 13 tests that set their own root to `HARNESS_TMP_ROOT` (required, not cosmetic: `TMP_ROOT` reads it first).
- Verify the resolved Chromium binary, not just the package, using the same resolution as `playwright-mcp.sh`.
- Offset the e2e slot ports by 20 so no slot reuses the ports `config.e2e.toml` declares, which belong to a human `make start-e2e` that does not take part in the slot lease.
- Resolve the orphan flag through `reviews.mjs`, the source of truth `block-merger-without-review` already uses.
- Surface the e2e verdict on the `#technical-harness` board, which could otherwise show every ticket green while the suite was red.
- Stop `cleanup-session` deleting `$REPO/test-results`: it is not session-scoped, breaking that hook's own invariant, and it is now vestigial since the suite runs in the `_session` worktree.
- Tests for the five uncovered guards. Every hook now has one.

Two hook comments were corrected rather than their code, because in both cases the code was right: `restrict-documentator-bash` cannot fail closed on an unknowable identity (it runs on every caller's Bash), and `block-orchestrator-merge` over-matches prose deliberately, since it is opt-in and fails loud while a cleverer regex would eventually let a real merge through. Both are now pinned by tests.

No application code is touched: harness hooks, agent prompts, rules, tests and config only.

## How to test

```bash
npm run test:unit:claude
```

382 tests over 41 files, green. Then check the two behaviors that were invisible before:

1. **No `/tmp` leak.** `ls /tmp | wc -l` before and after the run: unchanged. On `main` the same run leaves four new directories, including `/tmp/<sanitized-repo>/<session-id>/hooks.log`.
2. **The MCP browser.** `make install-playwright-browsers`, then reload the window and call any `mcp__playwright__browser_*` tool. Before this branch, `ensure-playwright-mcp` reported "present" and the first navigate failed with `Chromium distribution 'chrome' is not found at /opt/google/chrome/chrome`.

The e2e leg itself: `E2E_SMOKE_DRY=1 bash .claude/scripts/e2e-smoke.sh` prints the resolved slot, ports and rendered config without starting Docker. Slot 0 should now report `api_port=54361 app_port=5176`, clear of the dev stack (54321-54329) and the human e2e stack (54340-54349, 5175).


registry.json
```diff
@@ -481,6 +481,10 @@
           "path": "src/components/atomic-crm/misc/ActiveFilterButton.tsx",
           "type": "registry:component"
         },
+        {
+          "path": "src/components/atomic-crm/login/authConfig.ts",
+          "type": "registry:component"
+        },
```
erwanMarmelab, line 491: Not necessary in this PR
```suggestion
```

.claude/hooks/e2e-on-feature-review.mjs
erwanMarmelab, line 1: `settings.json` invokes it as a bare path with no `node` prefix, so the shell needs the exec
bit. Reproduced on a sibling 644 hook: `sh -c '<path>'` gives `Permission denied` (exit 126).

The hook never runs, so the suite is never launched, `e2e-result.json` is never written, and
`completion-invariant` never fires. The tests miss it because they spawn `node [HOOK]`
explicitly.

    git update-index --chmod=+x .claude/hooks/e2e-on-feature-review.mjs

.claude/agents/quality-reviewer.md
```diff
@@ -88,6 +88,21 @@ mistake would be most costly. Rules:
...
 
+**Record your verdict flag (required, do this BEFORE emitting the contract line).** The
+end-of-feature e2e suite is launched by a hook on your stop, and it keys off this flag, so a
+`BLOCKED` review never pays for a 10-minute suite. Write it yourself with a single Bash call:
+a post-stop transcript read races the flush and silently drops the verdict. The flag dir is
+`${TICKETS_DIR}/reviews`, and the key is the literal `FEATURE` (this pass has no `TASK_ID`):
+
+- APPROVED:
+  ```bash
+  RD="${TICKETS_DIR}/reviews" && mkdir -p "$RD" && touch "$RD/FEATURE-quality-reviewer"
```
erwanMarmelab: `RD="${TICKETS_DIR}/reviews"` relies on an env var nothing sets — `settings.json`'s `env` block has two keys and neither is this one (`completion-invariant.mjs:73` says so explicitly). Run verbatim by the agent it expands to `mkdir -p /reviews` → fails → the `touch` never runs → `FEATURE-quality-reviewer` never exists → the suite never launches even on APPROVED. The per-ticket block a few lines up says "substitute the literal value from your spawn prompt"; this one doesn't. Same note needed here (and `record-review-verdict` can't cover it — it requires a `TASK-\d+` key).

.claude/hooks/completion-invariant.mjs
```diff
@@ -95,6 +110,57 @@ try {
...
+    const p = join(
+      process.env.CHAT_SESSION_DIR || ctx.sessionDir,
+      "e2e-result.json",
+    );
+    if (!existsSync(p)) return; // not run this round -> nothing to react to
+    result = JSON.parse(readFileSync(p, "utf8"));
+  } catch {
+    return; // unreadable / malformed -> fail-open, same as the rest of this hook
+  }
+  if (result?.status !== "failed") return;
```
erwanMarmelab, line 131: `clearE2eRejects()` is only reached on the give-up path. After one red-e2e rejection the counter stays at 1 for the rest of the session, so the second red suite in that session is accepted with no rejection. Same "invariant silently never fires" class the PR fixes elsewhere.

.claude/hooks/e2e-on-feature-review.mjs
```diff
@@ -0,0 +1,128 @@
...
+
+appendProgress(
+  ctx.sessionDir,
+  "[e2e] suite starting on the session worktree...",
+);
+
+const r = bash(`E2E_SMOKE_SRC='${src}' bash '${script}' 2>&1`, {
+  cwd: ctx.repo,
+  timeout: E2E_TIMEOUT_MS,
+});
```
erwanMarmelab:
```suggestion
const r = bash(`E2E_SMOKE_SRC='${src}' bash '${script}' 2>&1`, {
  cwd: ctx.repo,
  timeout: E2E_TIMEOUT_MS,
  // A 15-minute suite's output overruns spawnSync's 1 MB default, which SIGKILLs it
  // mid-run and records a false `failed` with innocuous-looking tail output.
  maxBuffer: 64 * 1024 * 1024,
});
```

.claude/hooks/bash-guard.mjs
```diff
@@ -107,7 +98,8 @@ const runsUnitTests = (c) =>
   /(npm\s+run\s+test(:unit)?(:[a-z]+)?|npm\s+test\b|npx\s+vitest|make\s+test(-unit)?(-[a-z]+)?)/.test(
     c,
   );
-const runsE2eTests = (c) => /(npx\s+playwright\s+test|make\s+test-e2e)/.test(c);
+const runsE2eTests = (c) =>
+  /(npx\s+playwright\s+test|make\s+test-e2e|e2e-smoke\.sh)/.test(c);
```
erwanMarmelab:
```suggestion
const runsE2eTests = (c) =>
  !/E2E_SMOKE_DRY=1/.test(c) &&
  /(npx\s+playwright\s+test|make\s+test-e2e|e2e-smoke\.sh)/.test(c);
```

Adding `e2e-smoke.sh` to the predicate, combined with the any-caller rule above, means this PR blocks both commands its own "How to test" section asks a human to run: `E2E_SMOKE_DRY=1` bash `.claude/scripts/e2e-smoke.sh` and `make test-e2e-ci`.

The suggestion exempts the dry run only — it starts no Docker stack and takes no slot lease, so none of what the rule protects is at stake.

`make test-e2e-ci` is a genuine call to make: either exempt the main session, or fix the docs. The comment above the rule says a main-session Bash call must be refused too, so I've left that one alone.

erwanMarmelab (changes requested):

erwanMarmelab (approved):

PR #342, opened 2026-08-04
Keep the contact CSV import alive when its dialog closes

pysnooLab:
## Problem

Switching to another browser tab during a large CSV import closes the import dialog while the import keeps running in the background.

The import state lived in the dialog, itself rendered by `ContactImportButton`. On tab refocus React Query refetches the list, `ContactListLayoutDesktop` swaps `ContactEmpty` (which renders that button) for the populated list, and the dialog unmounts. The batch loop survives as a closure, but its state updates go nowhere: no progress, no completion signal.

## Solution

- Import state moves to a `ContactImportProvider` mounted above anything that can unmount it. It renders the dialog once; the button is just a trigger.
- `CRM` wraps the resolved layout with `withContactImport` rather than each layout mounting the provider itself. `Layout` and `MobileLayout` stay agnostic (no app-wide module knows about contacts), and an app passing its own `layout` prop inherits the provider instead of crashing on a missing context. The wrapped layout is passed after the props spread, otherwise a caller-supplied `layout` would override it.
- The dialog is closable while running: Close is no longer disabled, and closing no longer calls `reset()`, which aborted the import via `importIdRef`.
- A sonner toast shows a progress bar while the dialog is closed, pinned to `bottom-right` so app notifications (`bottom-center`) cannot render in front of it. It carries a Stop action, since the dialog holding the other stop link is now closable.
- The import notifies on completion and on failure, dialog open or not.
- A `beforeunload` listener, active only while importing, makes the browser confirm a reload or tab close. The import stays client side, so confirming still kills it.

Known limit: `CRM` swaps `DesktopAdmin` and `MobileAdmin` on the `useIsMobile` breakpoint, which remounts the whole admin tree, so crossing that breakpoint mid-import still loses the import.

## How to test

```bash
npm run dev   # or npm run dev:demo for FakeRest
```

Generate a few thousand contacts in a CSV (header from `contacts_export.csv`), then Contacts &gt; Import CSV:

1. Start the import, switch to another tab, come back: the dialog is still open and still counting.
2. Close the dialog: a progress bar appears bottom-right, the list stays usable.
3. Keep browsing (dashboard, companies): the progress bar follows you.
4. Hit Stop on the progress bar: the import halts.
5. Let it finish: a notification reports the imported and error counts, the list refreshes.
6. Hit Ctrl+R while it runs: the browser asks to confirm.

`npm run test` covers these in `ContactImport.test.tsx`, including the original regression (the import survives the unmount of the subtree holding the button).


pysnooLab:
[contacts-3.csv](https://github.com/user-attachments/files/30734277/contacts-3.csv)
[contacts-1000.csv](https://github.com/user-attachments/files/30734279/contacts-1000.csv)

Files for testing

src/components/atomic-crm/contacts/ContactImportDialog.tsx
```diff
@@ -0,0 +1,179 @@
...
+import { FileField } from "@/components/admin/file-field";
+
+import { useContactImportContext } from "./ContactImportContext";
+import * as sampleCsv from "./contacts_export.csv?raw";
+
+const SAMPLE_URL = `data:text/csv;name=crm_contacts_sample.csv;charset=utf-8,${encodeURIComponent(
+  sampleCsv.default,
+)}`;
+
+export function ContactImportDialog() {
```
fzaninotto: I think the current import dialog is misleading: it both tells the user not to close it, and that they can close it. IMO, it's useless as the snackbar does the job of informing of the progression of the import. 

image

One thing to be noted, is that navigating to the deals page for the first time kills an ongoing import (because this is a lazy route). I don't know if this can be fixed.

src/components/atomic-crm/contacts/withContactImport.tsx
```diff
@@ -0,0 +1,11 @@
+import type { LayoutComponent } from "ra-core";
+
+import { ContactImportProvider } from "./ContactImportProvider";
+
+export const withContactImport =
```
fzaninotto: No need for this custom wrapper IMO, just put the provider in the default CRM layout. You already throw a meaningful message in the import button if the context is not found, so developers using a custom layout will be notified. It will also avoid the useMemo in CRM.

src/components/atomic-crm/contacts/ContactImportProvider.tsx
```diff
@@ -0,0 +1,90 @@
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import type { ReactNode } from "react";
+import { useNotify, useRefresh } from "ra-core";
+
+import { usePapaParse } from "../misc/usePapaParse";
+import { ContactImportContext } from "./ContactImportContext";
+import { ContactImportDialog } from "./ContactImportDialog";
```
fzaninotto: Avoid extracting components that are only used in this component. As we use a shadcn registry to distribute / update Atomic CRM, the less files we maintain, the better.

fzaninotto (changes requested):

src/components/atomic-crm/contacts/ContactImportProvider.tsx
```diff
@@ -0,0 +1,302 @@
...
+  const refresh = useRefresh();
+  const processBatch = useContactImport();
+  const { importer, parseCsv, reset } = usePapaParse<ContactImportSchema>({
+    batchSize: 10,
+    processBatch,
+  });
+  const [isDialogOpen, setIsDialogOpen] = useState(false);
+
+  const importerRef = useRef(importer);
+  importerRef.current = importer;
```
fzaninotto: setting a ref during render is unsafe, cf https://react.dev/reference/react/useRef:

> Do not write or read ref.current during rendering, except for [initialization.](https://react.dev/reference/react/useRef#avoiding-recreating-the-ref-contents) This makes your component’s behavior unpredictable.

Also, the importer from papaparse is not supposed to be mutable, why do you need a ref for it? A comment would be welcome.

src/components/atomic-crm/contacts/ContactImportProvider.tsx
```diff
@@ -0,0 +1,302 @@
...
+
+  useEffect(() => {
+    if (importer.state === "complete") {
+      const finished = importerRef.current;
+      if (finished.state !== "complete") return;
```
fzaninotto: this I don't understand. The importer is finished but the importer ref isn't ? In which case can it happen?

src/components/atomic-crm/contacts/ContactImportProvider.tsx
```diff
@@ -0,0 +1,302 @@
...
+    hasProgress,
+    importCount,
+    rowCount,
+    errorCount,
+    remainingTime,
+    translate,
+    onStop,
+  ]);
+
+  useEffect(
```
fzaninotto: why another effect? You should return the cleanup function from the previous effect.

fzaninotto (changes requested):

src/components/atomic-crm/contacts/ContactImportProvider.tsx
pysnooLab: I need two effects because first one depends on importCount, it re-runs on every batch and progress bar will blink.

erwanMarmelab:
need rebase

src/components/atomic-crm/layout/MobileLayout.tsx
```diff
@@ -4,14 +4,15 @@ import { Skeleton } from "@/components/ui/skeleton";
...
+import { DataImportProvider } from "../dataImport/DataImportProvider";
 import { useConfigurationLoader } from "../root/useConfigurationLoader";
 import { MobileNavigation } from "./MobileNavigation";
 import { PullToRefresh } from "./PullToRefresh";
 
 export const MobileLayout = ({ children }: { children: ReactNode }) => {
   useConfigurationLoader();
   return (
-    <>
+    <DataImportProvider>
```
erwanMarmelab, line 15: `DataImportProvider` shouldn't live inside the `layout`. layout is a public prop of `CRMProps`, so any app passing its own layout renders no provider at all — and every DataImportButton (contacts/companies/deals lists, the three empty states, `DashboardStepper`) throws `useDataImportContext must be used inside a <DataImportProvider>`, taking those pages down entirely.

Having to patch `StoryWrapper` in the same way is the same smell: the provider is a requirement of the feature, not of the layout.

Suggestion — wrap the layout once in `CRM.tsx` so it holds whatever layout is passed:

```
const withDataImport = (LayoutComponent: LayoutComponent): LayoutComponent =>
  function DataImportLayout(props) {
    return (
      <DataImportProvider>
        <LayoutComponent {...props} />
      </DataImportProvider>
    );
  };
```

then in `DesktopAdmin` / `MobileAdmin`: `const layout = useMemo(() => withDataImport(props.layout ?? Layout), [props.layout]);` and drop it from `Layout`, `MobileLayout` and `StoryWrapper`.

src/components/atomic-crm/dataImport/DataImportProvider.tsx
```diff
@@ -0,0 +1,141 @@
...
+              ? resources.filter(({ name }) => name === pinnedName)
+              : resources
+          }
+          resource={resource}
+          onResourceChange={setResourceName}
+          onStart={startImport}
+          onClose={() => setIsDialogOpen(false)}
+        />
+      )}
+      <DataImportProgressToast importer={importer} onStop={reset} />
```
erwanMarmelab: `onStop={reset}` short-circuits the completion path: `reset()` sends the importer back to `idle`, so the `state === "complete"` effect never fires — no `refresh()`, no notification. Stop at 400/1000 and the 400 rows already created stay invisible until a manual reload, with zero feedback that anything landed.

The gap already existed behind the dialog's stop link on `main`, but that link is gone, so the toast is now the way to stop an import.

```
const stopImport = () => {
  const stopped = importer.state === "running" ? importer : null;
  reset();
  if (!stopped) return;
  refresh();
  notify("crm.data_import.complete", {
    type: "info",
    messageArgs: { importCount: stopped.importCount, errorCount: stopped.errorCount },
  });
};
```

(or a dedicated `crm.data_import.stopped` key if "complete" reads wrong for an interrupted run)

src/components/atomic-crm/dataImport/DataImportProgressToast.tsx
erwanMarmelab, line 1: The progress bar ignores `errorCount`, so an import with rejected rows tops out below 100% and never visually completes — while the ETA right underneath does account for both (`rowCount - importCount - errorCount, usePapaParse.tsx:106`). The two disagree on the same toast.

erwanMarmelab (changes requested):

src/components/atomic-crm/layout/MobileLayout.tsx
pysnooLab: It was built that way at first, but as you can see in the earlier comments, we chose to move it into `Layout`.

fzaninotto (approved):

PR #344, opened 2026-08-05
Let admins filter by any account manager

pysnooLab:
## Problem

The account manager filter only offered a "Me" toggle, so an admin could see every record but could not isolate the contacts, companies or deals of a given sales.

## Solution

- Admins get a searchable account manager picker: next to the "Me" toggle in the contact and company sidebars, in place of the only-mine switch on the deal board. Non-admins are unchanged.
- The gate reuses `canAccess("sales", "list")`, already admin-only. No new role, no database change.
- `AutocompleteInput` gains an opt-in `clearable` prop, used only here, so the filter can be undone.
- Along the way: full text search on `sales` (without it the picker's search filters nothing, and the Users list search was broken too), an accessible name on the trigger when the label is hidden, and the deal board switch no longer says "Only companies I manage".

## How to test

1. `make start`, sign in as an administrator.
2. On Contacts, Companies and Deals: pick a teammate in the account manager picker, the list narrows to their records. Click the cross, everything comes back. Repeat at mobile width, through the filter sheet.
3. Sign in as a non-admin: no picker, only the "Me" toggle and the only-mine switch.

image

image


src/components/atomic-crm/providers/supabase/dataProvider.ts
```diff
@@ -293,6 +293,9 @@ const lifeCycleCallbacks: ResourceCallbacks[] = [
   },
   {
     resource: "sales",
+    beforeGetList: async (params) => {
+      return applyFullTextSearch(["first_name", "last_name"])(params);
```
erwanMarmelab, line 297: We can add "email"

src/components/admin/autocomplete-input.tsx
```diff
@@ -179,6 +185,15 @@ export const AutocompleteInput = (
...
   );
 
+  const handleReset = useEvent(() => {
+    field.onChange("");
+    setFilterValue("");
+    if (isFromReference) {
+      setFilters(filterToQuery(""));
+    }
+    setOpen(false);
+  });
```
erwanMarmelab: This function is practically the same as `handleChange`. Wouldn't it be possible to use `handleChange` directly?

src/components/atomic-crm/deals/DealList.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+  description: "",
+  expected_closing_date: "2025-02-01T09:00:00.000Z",
+  id: 1,
+  index: 0,
+  name: "Acme deal",
+  sales_id: 0,
+  stage: "opportunity",
+  updated_at: "2025-01-01T09:00:00.000Z",
+  ...overrides,
+});
```
erwanMarmelab: Could you please move it to `StoryWrapper.tsx` with other builders?

erwanMarmelab (changes requested):

src/components/atomic-crm/providers/supabase/dataProvider.ts
pysnooLab: `sales` can't take `"email"` as-is. `applyFullTextSearch` maps the column
name `email` to `email_fts@ilike`, and `email_fts` only exists on the
`contacts_summary` view, where the email is a jsonb array that has to be
flattened first. `sales` is queried as a plain table and its `email` is a
scalar `citext`, so we would send `sales.email_fts@ilike` and PostgREST
would 400 on every search, both in the picker and in the Users list.

Making it work means dropping that special case and letting each caller
name its real column: `contacts` passes `email_fts` / `phone_fts`, `sales`
passes `email`. That removes the hidden mapping from the helper, and
`ilike` works fine on `citext`.

It does touch the contacts search, which is outside this PR's scope.

src/components/admin/autocomplete-input.tsx
```diff
@@ -81,6 +81,7 @@ export const AutocompleteInput = (
     Partial<Pick<InputProps, "source">> &
     ChoicesProps & {
       className?: string;
+      clearable?: boolean;
```
fzaninotto, line 84: this should be done in shadcn admin kit ;)

src/components/atomic-crm/contacts/ContactListFilter.tsx
```diff
@@ -150,6 +154,7 @@ export const ContactListFilter = () => {
           value={{ sales_id: identity?.id }}
           size={isMobile ? "lg" : undefined}
         />
+        {canFilterByAccountManager && <AccountManagerFilter />}
```
fzaninotto: I don't like the autocomplete in the sidebar. We're targeting SMEs, so there shouldn't be too many sales. I'd prefer if each sale was a selectable pill, as the other filters in the sidebar. We can limit the number of displayed sales to 10 to avoid breaking the layout on large teams. And even larger teams will need to change the UI to get a better fit for their use case. 
Also, you're actually displaying the "me" filter twice: once as an isolated filter, and once in the list of all sales. The best UI IMO would be:
1. me
2. Sales A
3. Sales B
4. ...
Without "me" ever being in the list of Sales

src/components/atomic-crm/sales/useCanFilterByAccountManager.ts
```diff
@@ -0,0 +1,9 @@
+import { useCanAccess } from "ra-core";
+
+export const useCanFilterByAccountManager = () => {
```
fzaninotto: I'd name it `canAccessSalesList`, and that way you don't have to alter the return value of `canAccess`. Not sure that it's necessary though, as it's a 2-liner. Less small files = less jumps when reading code = more understandable code.

fzaninotto (changes requested):

src/components/admin/autocomplete-input.tsx
pysnooLab: https://github.com/marmelab/shadcn-admin-kit/pull/173
https://github.com/marmelab/shadcn-admin-kit/pull/174

fzaninotto (approved):

PR #345, opened 2026-08-11
Attach forwarded emails sent from a sale's secondary address

pysnooLab:
## Problem

Sales who have more than one email address lose their forwarded contact discussions: the postmark webhook only matches the sender against `sales.email`, so a message forwarded from a secondary address is never attached.

## Solution

Add a `secondary_emails` jsonb column on `sales`, editable by each sale from the profile page (desktop) and the settings page (mobile). The postmark webhook falls back to `secondary_emails` when the sender matches no primary address. The `users` edge function trims, lowercases and deduplicates the addresses on write, rejects malformed ones and addresses already registered by another sale.

## How to test

1. `make start`, then add a secondary address in Profile (desktop) or Settings (mobile).
2. Send an inbound email whose `FromFull.Email` is that secondary address (see the curl samples at the bottom of `supabase/functions/postmark/index.ts`).
3. The note is attached to the contact with the right `sales_id`.

Automated: `make test` covers the lookup, the normalization and the error mapping.

## Hotspots

- `findEmailUsedByAnotherSale` builds its filter with `eq()`/`contains()`, never string interpolation: the values are user-controlled.
- An ambiguous sender (one address registered by two sales) is attributed to nobody and logged, rather than guessed.
- Uniqueness is enforced by the application, not by a constraint: jsonb cannot carry a unique index. Accepted trade-off, the residual risk is a concurrent save.
- `sales` keeps its `using (true)` select policy, so a secondary address is readable by any authenticated user, like the primary one already was. Deliberate for this PR, restricting it needs its own design since the profile page and every sales `ReferenceInput` read that table.


supabase/functions/postmark/index.ts
```diff
@@ -52,9 +52,23 @@ Deno.serve(async (req) => {
...
+
+  if (allSales.error) {
+    console.error("Could not fetch sales emails:", allSales.error);
+    return new Response("Could not fetch sales from database", { status: 500 });
+  }
+
   const salesEmails =
-    allSales.data?.map((s: { email: string }) => s.email) ?? [];
+    allSales.data
+      ?.flatMap((s: { email: string; secondary_emails: unknown }) => [
```
erwanMarmelab: why secondary_email is unknown ? should it be `string [] | null` ?

supabase/migrations/20260812093000_sales_secondary_emails_is_array.sql
erwanMarmelab, line 1: please merge your two migrations

supabase/schemas/01_tables.sql
```diff
@@ -97,7 +97,9 @@ create table public.sales (
     administrator boolean not null,
     user_id uuid not null,
     avatar jsonb,
-    disabled boolean not null default false
+    disabled boolean not null default false,
+    secondary_emails jsonb not null default '[]'::jsonb,
```
erwanMarmelab, line 101: new column sits on `public.sales`, whose only policy is `select to authenticated using (true)`. 

So every logged-in user can read colleagues' personal addresses. But UI hides them.

Need a column-level policy

supabase/functions/postmark/index.ts
```diff
@@ -127,6 +163,25 @@ Deno.serve(async (req) => {
...
+      failedContacts.push(email);
+    }
+  }
+
+  if (failedContacts.length) {
+    // A partial failure answers 403 so Postmark does not redeliver: the notes
+    // already written are not idempotent and would be duplicated. When every
+    // contact failed there is no note to duplicate, so a retry is safe.
+    // https://postmarkapp.com/developer/webhooks/inbound-webhook#errors-and-retries
+    const status = failedContacts.length === contacts.length ? 500 : 403;
```
erwanMarmelab, line 183: the new 500 makes Postmark redeliver, but attachments already uploaded at line 134. Each retry leaks a full copy. 

May be we can upload after contact resolution

erwanMarmelab (changes requested):
We don't check whether the new “secondary_email” address has already been entered as the primary email (“contacts.email”) for another record.

I don't think we need to check whether two “secondary_email” addresses are the same (in case of emails like “contact@marmelab.com”), but I do think we should check that.

pysnooLab:
> We don't check whether the new “secondary_email” address has already been entered as the primary email (“contacts.email”) for another record.
> 
> I don't think we need to check whether two “secondary_email” addresses are the same (in case of emails like “[contact@marmelab.com](mailto:contact@marmelab.com)”), but I do think we should check that.

sales.email is already covered: findEmailUsedByAnotherSale checks both columns for every address (isTakenBy("email", …) and isTakenBy("secondary_emails", …), excluding the sale being edited), so a secondary address that is another sale's primary one gets a 409 (email_taken). A separate check rejects an address equal to the sale's own primary email (secondary_email_is_primary). contacts.email is a different namespace, a contact and a sale can legitimately share an address.

supabase/schemas/01_tables.sql
pysnooLab: Postgres has no column-level RLS. Hiding only secondary_emails means either revoking the column grant on public.sales (today grant all on table public.sales to authenticated) and reading the table through a view, or moving the column to its own table with a user_id = auth.uid() policy. Both change every sales ReferenceInput plus the profile page, so it needs its own design. Note that sales.email is already readable by any authenticated user, so this PR doesn't widen the exposure class, it adds a row to a table that is already fully readable. I'd rather track it as a follow-up.

erwanMarmelab (approved):

PR #347, opened 2026-08-20
Fix supabase warnings and error

erwanMarmelab:
## Problem

Supabase studio mention us many warnings and one error

## Solution

Fix them

## How To Test

Apply the migrations and open [your local Supabase studio](http://localhost:54323/project/default/advisors/security?preset=WARN&id=)

## Additional Checks

- ~~[ ] Tested with **fakerest** provider (see [related documentation](https://github.com/marmelab/atomic-crm/blob/main/doc/developer/data-providers.md))~~


src/components/atomic-crm/providers/supabase/dataProvider.ts
```diff
@@ -402,13 +402,16 @@ const applyFullTextSearch = (columns: string[]) => (params: GetListParams) => {
...
     if (fi.path) {
-      const { error } = await getSupabaseClient()
+      const { data } = getSupabaseClient()
         .storage.from(ATTACHMENTS_BUCKET)
-        .createSignedUrl(fi.path, 60);
+        .getPublicUrl("");
 
-      if (!error) {
+      if (fi.src.startsWith(data.publicUrl)) {
         return fi;
```
pysnooLab: `rawFile` is absent on persisted attachments (they come back from JSONB), so
`file.name` on line 438 throws if we ever reach it. On main that was
unreachable: `createSignedUrl` always succeeded for an object still in the
bucket, so we returned early. 
Now the early return depends on `fi.src` matching
the *current* public URL base. If that base ever changes (custom domain, project
migration, a different `VITE_ATTACHMENTS_BUCKET`), the prefix stops matching,
`fetch(src)` returns 200 since the bucket is public, and saving the note dies on
`fi.rawFile.name`.

`fi.path` is only ever set by our own uploads (below, and in the postmark
function), so it already carries exactly the information we need:

```suggestion
    // Already an object of our bucket: nothing to re-upload. `path` is only set
    // by our own uploads, so this needs no read on storage.objects (a SELECT
    // policy there would let any signed-in user list every attachment).
    if (fi.path) {
      return fi;
    }
```

pysnooLab (changes requested):

pysnooLab (approved):

supabase/schemas/08_extensions.sql
```diff
@@ -0,0 +1,13 @@
...
+-- a second API surface this app never queries (it uses PostgREST only). It is
+-- what makes linters 0026 / 0027 fire on every table and view; those lints gate
+-- solely on the extension being installed. Reinstall with
+-- `create extension pg_graphql` to bring GraphQL back.
+--
+-- `supabase db diff` does not diff extensions, so this file documents the
+-- intent but does not enforce it; the drop lives in a hand-written migration.
+--
+
+drop extension if exists pg_graphql cascade;
```
fzaninotto: ### Could we revert the `pg_graphql` drop?

The rest of this PR is a solid hardening pass, but I'd like to keep GraphQL —
having a second API surface is a feature of the product, not an accident, and I
don't think the lints require dropping it.

**The lints are informational.** 0026 / 0027
(`pg_graphql_anon_table_exposed` / `pg_graphql_authenticated_table_exposed`) are
not misconfigurations. The docs are explicit:

> If introspection visibility is intentional for a relation, the lint can be
> safely ignored for that object.

They report what GraphQL introspection makes *discoverable* — names, columns,
relationships — while RLS still governs the data itself. `pg_graphql` is no more
permissive than PostgREST here: both are gated by the same GRANTs and the same
policies.

**The grant cleanup in this PR already fixes the half that matters.** Once `anon`
holds no grants on `public`, 0026 stops firing on its own. What remains is 0027,
i.e. tables visible to authenticated users — which is precisely the intended
behaviour for a GraphQL API we want to keep.

So dropping the extension removes a working feature to get a fully green
Advisor, and I'd rather keep the feature and accept the informational warning.

fzaninotto (changes requested):

PR #348, opened 2026-08-20
feat: Import data from CSV

erwanMarmelab:
## Problem

We can't import data from CSV in the onboarding page.

## Solution

- Add an "Import data" button in the onboarding page
- Remove the "Import CSV" button for contacts
- Fit the styles for the "Add note" and the "Add contact" buttons
- Add the new "Import data" button on every resources

## Additional Checks


src/components/atomic-crm/dataImport/DataImportDialog.tsx
```diff
@@ -11,70 +17,68 @@ import {
...
   onClose,
-}: ContactImportModalProps) {
+  resources,
+}: DataImportDialogProps) {
   const translate = useTranslate();
+  const getResourceLabel = useGetResourceLabel();
   const refresh = useRefresh();
-  const processBatch = useContactImport();
-  const { importer, parseCsv, reset } = usePapaParse<ContactImportSchema>({
+  const [resource, setResource] = useState(resources[0]);
```
pysnooLab: `useState(resources[0])` freezes the resource object and its `processBatch` at mount, so
the closure keeps whatever `useConfigurationContext()` returned on the first render.
`useConfigurationLoader` fills that store asynchronously from the backend, so on a first
load the closure stays pinned to `defaultConfiguration`.

On a tenant with a customized pipeline (stages qualification/negotiation, category
retainer), importing `Deal A,Negotiation,Retainer` stores
`stage=opportunity category=undefined`, while reading the hook at click time stores
`stage=negotiation category=retainer`. A valid stage and category from the CSV are
discarded and the deal is written with a stage that does not exist in that pipeline, so
stage filters and reports are wrong. Same for `sector` in `useCompanyImport`.

Reading the resource from `resources` on each render (keeping only the selected name in
state) fixes it.

src/components/atomic-crm/dataImport/useDealImport.ts
```diff
@@ -0,0 +1,63 @@
...
+              category: toConfiguredValue(row.category, dealCategories),
+              // stage is required, so fall back to the first configured one
+              stage:
+                toConfiguredValue(row.stage, dealStages) ??
+                dealStages[0]?.value,
+              description: toText(row.description),
+              amount: toNumber(row.amount),
+              expected_closing_date: toIsoDate(row.expected_closing_date),
+              sales_id: identity?.id,
+              index: 0,
```
pysnooLab: All imported deals get `index: 0`, without the same-stage `index + 1` shift that
`DealCreate.onSuccess` applies.

`onDragEnd` passes the destination *deal* to `updateDealStage`, and only its `index`
field is read. With every index at 0, dragging a card emits a single write,
`{id, index: 0}`, so the drop persists nothing and `refetch()` snaps the card back.
Only dropping after the last card works. Column order also becomes whatever row order
Postgres happens to return, since `getDealsByStage` sorts on equal keys.

src/components/atomic-crm/dataImport/useCompanyImport.ts
```diff
@@ -0,0 +1,56 @@
...
+ * left empty — except `name`, which the database requires.
+ */
+export function useCompanyImport(): ProcessImportBatch {
+  const { companySectors } = useConfigurationContext();
+  const { identity } = useGetIdentity();
+  const dataProvider = useDataProvider();
+
+  return useCallback(
+    async (batch) => {
+      await Promise.all(
```
pysnooLab: `Promise.all` rejects on the first bad row, but the other creates already fired. In
`usePapaParse` a rejected batch adds `batch.length` to `errorCount` and 0 to
`importCount`, so the records exist while the report says they do not.

A 3-row CSV with one empty `name` (NOT NULL) creates 2 companies and reports
"Import complete. Imported 0 records, with 3 errors". Users will fix the file and
re-import, producing duplicates. Same shape in `useDealImport.ts:35`.

src/components/atomic-crm/dataImport/useDealImport.ts
```diff
@@ -0,0 +1,63 @@
...
+                ? companies.get(companyName)?.id
+                : undefined,
+              contact_ids: [],
+              category: toConfiguredValue(row.category, dealCategories),
+              // stage is required, so fall back to the first configured one
+              stage:
+                toConfiguredValue(row.stage, dealStages) ??
+                dealStages[0]?.value,
+              description: toText(row.description),
+              amount: toNumber(row.amount),
```
pysnooLab: `amount` is not rounded, but the column is `bigint`. PostgREST rejects the row:
`22P02 invalid input syntax for type bigint: "4500.50"`. Combined with the batch
accounting above, a single amount with cents reports its whole batch of 10 as failed.

src/components/atomic-crm/dataImport/useCompanyImport.ts
```diff
@@ -0,0 +1,56 @@
...
+          dataProvider.create("companies", {
+            data: {
+              name: toText(row.name),
+              sector: toConfiguredValue(row.sector, companySectors),
+              size: sizeOf(row.size),
+              linkedin_url: toText(row.linkedin_url),
+              website: toText(row.website),
+              phone_number: toText(row.phone_number),
+              address: toText(row.address),
+              zipcode: toText(row.zipcode),
```
pysnooLab, line 32: `usePapaParse` runs with `dynamicTyping: true`, so these text columns lose leading
zeros: `02134` is stored as `2134`, `0155123456` as `155123456`, and a zero-prefixed
`tax_identifier` is truncated. The insert succeeds silently because Postgres casts the
number into the `text` column.

Pre-existing papaparse option, but newly exposed on zipcode / phone_number /
tax_identifier. The sample CSV dodges it (10001, 555-0100), so it only shows with real
data.

src/components/atomic-crm/dataImport/parseCell.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (text === undefined) return undefined;
+  const value = Number(text);
+  return Number.isFinite(value) ? value : undefined;
+};
+
+/** Cell content as an ISO 8601 date, or undefined when it is empty or invalid. */
+export const toIsoDate = (cell: ImportCell): string | undefined => {
+  const text = toText(cell);
+  if (text === undefined) return undefined;
+  const date = new Date(text);
```
pysnooLab: `new Date(text)` parses non-ISO dates as local midnight. In a UTC+ timezone,
`09/30/2026` and `2026-09-30T00:00:00` both become `2026-09-29T22:00:00Z`, which this
`date` column truncates to 2026-09-29, one day early. And `30/09/2026` returns
undefined, so the closing date is dropped without counting as an error.

ISO date-only input (the documented format) is fine. Note `dealUtils.formatISODateString`
already refuses `new Date(isoString)` for this exact reason on the read path.

src/components/atomic-crm/dashboard/DashboardStepper.tsx
```diff
@@ -88,22 +93,17 @@ export const DashboardStepper = ({
...
                     ) : (
-                      <>
-                        <CreateButton
-                          label="resources.contacts.action.new"
-                          resource="contacts"
-                        />
-                        <ContactImportButton />
-                      </>
+                      <Button asChild className="w-fit">
+                        <Link role="button" to="/contacts/create">
```
pysnooLab: `role="button"` on an anchor: it is announced as a button but Space does not activate it,
and the link affordances are lost. `CreateButton` renders a plain `<Link>` with no role,
so this deviates from the codebase convention. The hardcoded path instead of
`useCreatePath()` is equivalent at runtime, but same remark.

Also, `DashboardStepper.test.tsx` runs at the default viewport, which is below the 768px
breakpoint, so it exercises the mobile `<button>` branch and this desktop path is untested.

src/components/atomic-crm/misc/useImportFromJson.ts
```diff
@@ -806,10 +807,3 @@ const isTask = (data: any): data is TaskImport =>
  * Maps a company size number to the appropriate size category.
  * Categories: 1, 10, 50, 250, 500
  */
-const mapSizeToCategory = (size: number): 1 | 10 | 50 | 250 | 500 => {
```
pysnooLab, line 809: Orphan JSDoc left behind when `mapSizeToCategory` moved to `sizes.ts`. The file now ends
on a comment with no declaration under it.

src/components/atomic-crm/dataImport/DataImportButton.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+const listAll = (dataProvider: DataProvider, resource: string) =>
+  dataProvider.getList(resource, {
+    filter: {},
+    pagination: { page: 1, perPage: 25 },
+    sort: { field: "id", order: "ASC" },
+  });
+
+/** Imports one batch on click, so a test controls when the import starts. */
+const ImportHarness = ({
```
pysnooLab, line 28: `ImportHarness` calls the import hooks directly, which bypasses the dialog wiring where
the `sales_id` bug lives, and no test asserts `sales_id`. One test driving the real
dialog end to end would cover both.

src/components/atomic-crm/dataImport/DataImportDialog.tsx
```diff
@@ -190,11 +216,11 @@ export function ContactImportDialog({
           </div>
         </Form>
 
-        <div className="flex justify-start pt-6">
+        <div className="flex justify-start">
           <FormToolbar>
             {importer.state === "idle" ? (
               <Button onClick={startImport} disabled={!file}>
-                {translate("resources.contacts.import.button")}
+                {translate("crm.data_import.button")}
```
pysnooLab: nit : This submit button reuses `crm.data_import.button`, the same label as the trigger on the
page, so two "Import CSV" controls coexist on screen. A distinct key would help both
users and tests.

pysnooLab (changes requested):

src/components/atomic-crm/dataImport/DataImportDialog.tsx
erwanMarmelab: > so two "Import CSV" controls coexist on screen

On which page do you see these two “Import CSV” options??

src/components/atomic-crm/dataImport/useImportableResources.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+export function useImportableResources(): ImportableResource[] {
+  const processContacts = useContactImport();
+  const processCompanies = useCompanyImport();
+  const processDeals = useDealImport();
+  const definitions = useResourceDefinitions();
+
+  const resources: ImportableResource[] = [
+    {
+      name: "contacts",
```
pysnooLab, line 30: The contacts entry declares no textColumns, so dynamicTyping still applies to all of its columns: phone_work,0155123456 lands in phone_jsonb as the number 155123456, while the type says { number: string }

pysnooLab (changes requested):

src/components/atomic-crm/dataImport/DataImportDialog.tsx
pysnooLab: Not two buttons on the page: the list button stays visible behind the dialog, and the submit button carried the same label. But you already fixed with crm.data_import.start

pysnooLab (approved):

PR #349, opened 2026-08-21
feat(Mobile): Pull to refresh

erwanMarmelab:
## Problem

No refresh button on mobile UI

ex: checking the notes tab on a contact to see the latest note forwarded to postmark

## Solution

Pull to refresh : a universal listener that detects when the user starts a scroll up from the top and shows a refresh button, then refreshes the data.

## Screencast

https://github.com/user-attachments/assets/adab642a-3e00-4b69-91e7-96ecce657f99

## Additional Checks

### DIFF
--- a/registry.json
+++ b/registry.json
@@ -510,11 +510,19 @@
           "type": "registry:component"
         },
         {
+          "path": "src/components/atomic-crm/layout/useRefreshData.ts",
+          "type": "registry:component"
+        },
+        {
           "path": "src/components/atomic-crm/layout/TopToolbar.tsx",
           "type": "registry:component"
         },
         {
           "path": "src/components/atomic-crm/layout/PullToRefresh.tsx",
+          "type": "registry:component"
+        },
+        {
+          "path": "src/components/atomic-crm/layout/MobileRefreshButton.tsx",
           "type": "registry:component"
         },
         {
--- a/registry.json
+++ b/registry.json
@@ -482,6 +482,10 @@
           "type": "registry:component"
         },
         {
+          "path": "src/components/atomic-crm/login/authConfig.ts",
+          "type": "registry:component"
+        },
+        {
           "path": "src/components/atomic-crm/login/StartPage.tsx",
           "type": "registry:component"
         },
@@ -507,6 +511,10 @@
         },
         {
           "path": "src/components/atomic-crm/layout/TopToolbar.tsx",
+          "type": "registry:component"
+        },
+        {
+          "path": "src/components/atomic-crm/layout/PullToRefresh.tsx",
           "type": "registry:component"
         },
         {
--- a/src/components/atomic-crm/layout/PullToRefresh.tsx
+++ b/src/components/atomic-crm/layout/PullToRefresh.tsx
@@ -1,7 +1,7 @@
-import { useQueryClient } from "@tanstack/react-query";
 import { LoaderCircle, RotateCw } from "lucide-react";
-import { useTranslate } from "ra-core";
 import { useCallback, useEffect, useRef, useState } from "react";
+
+import { useRefreshData } from "./useRefreshData";
 
 /** Finger travel below which we cannot tell a pull from a tap, so we stay passive. */
 const GESTURE_SLOP = 8;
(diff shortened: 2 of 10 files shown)


src/components/atomic-crm/layout/PullToRefresh.tsx
```diff
@@ -0,0 +1,166 @@
...
+    } finally {
+      isRefreshingRef.current = false;
+      setIsRefreshing(false);
+      moveTo(0);
+    }
+  }, [moveTo, queryClient]);
+
+  useEffect(() => {
+    const onTouchStart = (event: TouchEvent) => {
+      startYRef.current =
```
pysnooLab: A second finger sets startYRef to null, and onTouchEnd (L108) early-returns without resetting distance. Reproduced on Pixel 5 with real touch: 3s after both fingers lift the button is still at translateY(96px), and since distanceRef is still 96, the next tap anywhere fires a full invalidateQueries().


if (event.touches.length !== 1 || isRefreshingRef.current || isScrollGesture(event.target)) {
  startYRef.current = null;
  if (!isRefreshingRef.current) moveTo(0);
  return;
}
startYRef.current = event.touches[0].clientY;

src/components/atomic-crm/layout/PullToRefresh.tsx
```diff
@@ -0,0 +1,166 @@
...
+const MIN_SPINNER_MS = 400;
+
+/**
+ * True when a pull starting on `target` should scroll the page rather than refresh it:
+ * some ancestor is already scrolled down, or the touch is inside a dialog/sheet, which
+ * owns its own scrolling.
+ */
+const isScrollGesture = (target: EventTarget | null) => {
+  let node = target instanceof Element ? target : null;
+  if (node?.closest('[role="dialog"], [data-vaul-drawer]')) {
```
pysnooLab: [role="dialog"] misses Radix Select (role="listbox") and DropdownMenu (role="menu"), both portalled to body so they are not inside their sheet either. Reproduced: pulling inside an open task-actions menu refreshes the whole app with the menu still open. Suggest detecting the nearest scrollable ancestor instead of allowlisting roles. [data-vaul-drawer] is redundant, vaul already renders role="dialog".

src/components/atomic-crm/layout/PullToRefresh.tsx
```diff
@@ -0,0 +1,166 @@
...
+        return;
+      }
+      const delta = event.touches[0].clientY - startYRef.current;
+      if (delta <= GESTURE_SLOP) {
+        moveTo(0);
+        return;
+      }
+      // We own the gesture from here: suppress the native overscroll, which on Chrome
+      // for Android is its own pull-to-refresh reloading the whole page.
+      event.preventDefault();
```
pysnooLab: Per Chrome's own docs, the overscroll-behavior-y: contain you added is already enough to kill Android's native pull-to-refresh, and none additionally kills the glow and the iOS rubber-band. If index.css uses none, this preventDefault() can go and the listener at L125 can be passive: true.

As it stands it cancels legitimate scrolling: on a diagonal swipe over a horizontally scrollable area. It also keeps a non-passive touchmove on document for every mobile page.

src/components/atomic-crm/layout/PullToRefresh.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  const progress = Math.min(1, distance / TRIGGER_DISTANCE);
+
+  return (
+    <div className="fixed inset-x-0 top-0 z-50 flex justify-center pointer-events-none">
+      <button
+        type="button"
+        aria-label={translate("ra.action.refresh")}
+        disabled={isRefreshing}
+        onClick={() => void refresh()}
```
pysnooLab: onClick is not reachable in real use. Verified with a capture-phase click listener: pulling under the trigger detaches the button before the click is synthesized (0 clicks), and during a refresh it is disabled (0 clicks). The only state where it is visible and enabled is the multi-touch bug above. Drop the handler, or make the affordance persistent.

src/index.css
```diff
@@ -186,3 +186,12 @@ input[type="time"] {
 body {
   font-family: "Inter Variable", ui-sans-serif, system-ui, sans-serif;
 }
+
+/* The mobile pull-to-refresh (see layout/PullToRefresh.tsx) owns the overscroll
+   gesture: without this, Chrome for Android runs its own pull-to-refresh — a full
+   page reload — on top of ours. 767px is the mobile breakpoint of useIsMobile(). */
+@media (max-width: 767px) {
+  html {
+    overscroll-behavior-y: contain;
```
pysnooLab: Make it none rather than contain if the preventDefault() goes, so the CSS also covers the iOS rubber-band.

registry.json
```diff
@@ -481,6 +481,10 @@
           "path": "src/components/atomic-crm/misc/ActiveFilterButton.tsx",
           "type": "registry:component"
         },
+        {
```
pysnooLab, line 484: Not needed for this PR, but redundant everywhere because it was not added

pysnooLab (changes requested):
Docs: the gesture is undiscoverable, and doc/src/content/docs/users/mobile-app.mdx has a Usage section right above the Offline Mode note that describes this exact problem.

Accessibility: no non-gesture way to refresh on mobile now. A button in MobileHeader would fix that and justify the onClick.

src/components/atomic-crm/layout/PullToRefresh.tsx
```diff
@@ -0,0 +1,177 @@
...
+const MAX_DISTANCE = 96;
+
+/** True when the element scrolls on its own and has somewhere to scroll to. */
+const isScroller = (node: Element) => {
+  const { overflowY } = getComputedStyle(node);
+  return (
+    (overflowY === "auto" || overflowY === "scroll") &&
+    node.scrollHeight > node.clientHeight
+  );
+};
```
pysnooLab: Don't think isScroller is mandatory. Scroll guard is already covered by `scrollTop > 0`
Deleting isScroller also removes a getComputedStyle per ancestor on every touchstart.

pysnooLab (approved):
Minor no action needed: the index.css rule applies to every page under 767px, including the login and signup routes, which render outside MobileLayout and so have no pull-to-refresh to replace the browser's own.

PR #350, opened 2026-08-21
feat: Come back on the filled form on task post fail

erwanMarmelab:
## Problem

When posting a task on destop from the contact view, the dialog closes immediately while the request is sent. If the request fails, an error message appears, but the user has to reopen the dialog and reenter the details, which are lost.

## Solution

Reopen the dialog on error and prefill with the task info

## How To Test

in your console write this:
```
const realFetch = window.fetch;
window.fetch = async (input, init) => {
  const url = typeof input === "string" ? input : input.url;
  if (init?.method === "POST" && url.includes("/rest/v1/tasks")) {
    await new Promise((r) => setTimeout(r, 5000));
    return new Response('{"message":"Boom"}', {
      status: 500,
      headers: { "Content-Type": "application/json" },
    });
  }
  return realFetch(input, init);
};
```

to rollback:
```
window.fetch = realFetch
```

## Additional Checks

- ~~[ ] Tested with **fakerest** provider (see [related documentation](https://github.com/marmelab/atomic-crm/blob/main/doc/developer/data-providers.md))~~
- [ ] Tested with **Mobile** resolution


pysnooLab (changes requested):
## The described problem does not reproduce

On `main`, the dialog stays open during the request and after a failed create, with the input intact. 

## It introduces a regression

Anything typed after clicking Save is wiped when the error lands.

## The real bug is on the success path, and this PR already half fixes it

When the task create succeeds but the follow-up contact calls fail:

| Failing call | Desktop (`main`) | Mobile (`main`) |
|---|---|---|
| `getOne("contacts")` | dialog closed, **no message** | sheet **stays open**, **no message** |
| `update(last_seen)` | closed, "Task added" | closed, "Task added" |

The task is created in every case. On mobile the filled sheet stays there, so the user clicks Save again and gets a duplicate. Reopening a prefilled form would make that worse, not better.

## Requested changes

1. Keep the `handleSuccess` hardening (`notify` moved before the contact calls, `try/catch` around `last_seen`).
2. Drop `failedTask`, `handleSettled`, the conditional `record` and the `onSettled` option.
3. Apply the same ordering to `TaskCreateSheet.tsx` (notify + close first, then best-effort `last_seen`). This also fixes the `if (!referenceRecordId) return` path, where the sheet currently never closes.
4. Add the missing test: `getOne("contacts")` fails after a successful create, assert the dialog is closed and "Task added" is shown, on both screens.
5. Move the unrelated `registry.json` entry to its own commit.

src/components/atomic-crm/tasks/TaskCreateSheet.tsx
```diff
@@ -42,24 +42,31 @@ export const TaskCreateSheet = ({
...
+
+    try {
+      const { data: contact } = await dataProvider.getOne("contacts", {
+        id: referenceRecordId,
+      });
+      await update("contacts", {
+        id: referenceRecordId as unknown as Identifier,
+        data: { last_seen: new Date().toISOString() },
+        previousData: contact,
+      });
```
pysnooLab: useUpdate() without returnPromise: true returns mutation.mutate(...), i.e. undefined, so this await resolves immediately. The try/catch therefore only covers getOne: an update failure escapes it, and since this useUpdate() has no options, ra-core notifies nothing. A failing last_seen update is completely silent. Either pass { returnPromise: true } so the catch really covers both calls, or narrow the comment above to say only getOne is guarded.

src/components/atomic-crm/tasks/AddTask.tsx
```diff
@@ -48,20 +49,24 @@ export const AddTask = ({
   };
   const getContactRepresentation = useGetRecordRepresentation("contacts");
 
-  const handleSuccess = async (data: any) => {
+  const handleSuccess = async (data: Task) => {
+    // The task is created: close and notify before the best-effort follow-up
+    // calls, so a failure there cannot leave the dialog open without feedback.
```
pysnooLab: Remove comments, already in commit message

src/components/atomic-crm/tasks/AddTask.tsx
```diff
@@ -48,20 +49,24 @@ export const AddTask = ({
...
-      data: { last_seen: new Date().toISOString() },
-      previousData: contact.data,
-    });
-
     notify("resources.tasks.added");
+
+    try {
+      const { data: contact } = await dataProvider.getOne("contacts", {
+        id: data.contact_id,
+      });
```
pysnooLab, line 59: The previous if (!contact.data) return; guard is gone: contact.id is now dereferenced directly, so a provider returning { data: undefined } throws a TypeError that the new catch swallows and logs as "Could not update the contact last_seen date". Low risk with our providers, but it widens the silent path rather than fixing it.

src/components/atomic-crm/tasks/TaskCreateSheet.tsx
```diff
@@ -42,24 +42,31 @@ export const TaskCreateSheet = ({
...
+      queryClient.invalidateQueries({
+        queryKey: ["contacts", "getOne"],
+      });
```
pysnooLab: Because the update above does not actually await the PUT, this invalidation fires before the contact is written, so any refetch it triggers re-reads the stale record. useUpdate already invalidates on settled, so this block can simply go.

src/components/atomic-crm/tasks/TaskCreateSheet.tsx
```diff
@@ -42,24 +42,31 @@ export const TaskCreateSheet = ({
   if (!identity) return null;
 
   const handleSuccess = async (data: any) => {
```
pysnooLab: AddTask was tightened to data: Task in this commit, this one keeps any plus the as unknown as Identifier cast on L60. Typing it as Task removes both.

src/components/atomic-crm/tasks/TaskCreateSheet.test.tsx
```diff
@@ -105,4 +105,34 @@ describe("TaskCreateSheet", () => {
     expect(updatedContact.data.last_seen).not.toBe(originalLastSeen);
```
pysnooLab: This assertion has no deterministic wait for the contact PUT, it relies on the fake provider's 300 ms latency. Closing and notifying earlier shortens the margin further. Wrap it in expect.poll like the task assertion on L75-84.

pysnooLab (changes requested):
The last commit reverts the failedTask + onSettled approach and instead confirms the creation before the best-effort contact refresh. The fix itself is right and much simpler, but three things need settling:

Title and description no longer match the diff. The delivered change is "confirm task creation even when the contact refresh fails". The How-To-Test (500 on POST /rest/v1/tasks) does not exercise it either: on main, setOpen(false) already lives in onSuccess only and CreateBase is pessimistic, so the dialog already stayed open with the values preserved on a failed create.
The one case the reverted approach handled is now dropped: dismissing the dialog while the create is in flight destroys the form (Radix unmounts the portal content), so a later failure loses the input. That is a regression against 1c7d3d40, not against main, and dropping it is defensible, but say so in the description instead of removing the test silently.
NoteCreateSheet.tsx:49-64 has the same defect: notify / redirect / onOpenChange(false) all sit after the awaited getOne, behind an if (!contact) return. A failing refresh leaves the sheet open with a misleading error toast on a note that was created. Same three-line fix here,.

src/components/atomic-crm/tasks/TaskCreateSheet.tsx
```diff
@@ -36,30 +35,34 @@ export const TaskCreateSheet = ({
...
+        {
+          id: data.contact_id,
+          data: { last_seen: new Date().toISOString() },
+          previousData: contact,
+        },
+        { returnPromise: true },
+      );
+    } catch (error) {
+      console.error("Could not update the contact last_seen date", error);
+    }
```
pysnooLab, line 65: My previous comment was wrong, sorry. Pessimistic useUpdate never invalidates: useMutationWithMutationMode does that only in optimistic/undoable mode (useMutationWithMutationMode.ts:190-205), otherwise it just setQueryData with updatedAt: now, which marks the cached contact fresh and suppresses the refetch. And that invalidation was not about last_seen: it refreshed nb_tasks, which comes from contacts_summary (03_views.sql:124) and changes with the task INSERT. Mobile ContactShow reads record.nb_tasks (ContactShow.tsx:64), so [Mobile Chrome] userAddingATask now fails 3 times on expect(page.getByText("1 task")) (userAddingATask.spec.ts:73); it was green on 12fb2704. Put it back right after onOpenChange(false), since the count depends on the create, not on the best-effort PUT (the useQueryClient import has to come back too):


queryClient.invalidateQueries({ queryKey: ["contacts", "getOne"] });

src/components/atomic-crm/notes/NoteCreateSheet.tsx
```diff
@@ -44,24 +44,34 @@ export const NoteCreateSheet = ({
...
+        {
+          id: referenceRecordId as unknown as Identifier,
+          data: { last_seen: new Date().toISOString(), status: data.status },
+          previousData: contact,
+        },
+        { returnPromise: true },
+      );
+    } catch (error) {
+      console.error("Could not update the contact last_seen date", error);
+    }
```
pysnooLab, line 74: The happy path is covered on mobile by onboarding.spec.ts:73-87, but the failure path this commit fixes is not tested anywhere (no NoteCreateSheet.test.tsx). Both task paths got exactly that test: getOne("contacts") rejects after a successful create, assert the sheet is closed and "Note added" is shown.

src/components/atomic-crm/tasks/AddTask.tsx
```diff
@@ -48,20 +49,27 @@ export const AddTask = ({
...
+        {
+          id: contact.id,
+          data: { last_seen: new Date().toISOString() },
+          previousData: contact,
+        },
+        { returnPromise: true },
+      );
+    } catch (error) {
+      console.error("Could not update the contact last_seen date", error);
+    }
```
pysnooLab, line 72: Fourth copy of the same best-effort block (here, TaskCreateSheet.tsx:49-65, NoteCreateSheet.tsx:58-74, and NoteCreate.tsx:91-99 which still fires and forgets with no try/catch). A useTouchContactLastSeen() hook removes the divergence and gives the invalidation above a single home.

src/components/atomic-crm/notes/NoteCreateSheet.tsx
```diff
@@ -44,24 +44,34 @@ export const NoteCreateSheet = ({
   if (!identity) return null;
 
   const handleSuccess = async (data: any) => {
```
pysnooLab, line 46: Still data: any here, plus the as unknown as Identifier cast on L66, while both task files are typed now. Typing it removes both, and data.contact_id replaces data[foreignKeyMapping["contacts"]].

pysnooLab (changes requested):

fzaninotto (commented):
I'm not convinced this change is even necessary. The initial bug doesn't exist, and the one this PR fixes (an update fails on the contact after a creation succeeds on the task) is so unlikely that we shouldn't bother adding that much code and complexity to fix it. I'm in favour of closing this PR unmerged. WDYT?

erwanMarmelab:
> I'm not convinced this change is even necessary. The initial bug doesn't exist, and the one this PR fixes (an update fails on the contact after a creation succeeds on the task) is so unlikely that we shouldn't bother adding that much code and complexity to fix it. I'm in favour of closing this PR unmerged. WDYT?

I agree

PR #355, opened 2026-09-09
feat: Global search

erwanMarmelab:
## Problem

When a user looks for a past note, the only possible way to find it is by contact. If they remember only a term of the note but not the contact name, they’re obliged to check all contacts one by one

## Solution

Add a search engine for notes & other content

## How

A `search_index` view flattens the six sources into `(resource, record_id, title, subtitle, content)` — same `UNION ALL` shape as `activity_log`, `security_invoker = on` so RLS still applies. The frontend queries it like any resource, with a single `content@ilike` filter; `ra-data-postgrest` ANDs one `ilike` per word, so "Thomas TF1" matches both words

## Additional Checks


supabase/schemas/03_views.sql
```diff
@@ -132,3 +132,104 @@ select count(sub.id) as is_initialized
...
+from public.tasks t
+    left join public.contacts tco on tco.id = t.contact_id
+union all
+select
+    ('contactNote.' || cn.id) as id,
+    'contact_notes' as resource,
+    cn.id as record_id,
+    cn.text as title,
+    concat_ws(' ', nco.first_name, nco.last_name) as subtitle,
+    concat_ws(
```
pysnooLab, line 215: **Displayed name is not searchable.** `subtitle` carries the parent contact/deal name, `content` does not, and the provider ANDs one `ilike` per word. "sourdough Thomas" returns 0 for a note on Thomas Anderson containing "sourdough", the exact use case this PR targets. Same in `deal_notes` (L229), `tasks` (L193) and `companies` (L143, `sector`).

Fold the subtitle expression into each arm's `content`.

supabase/schemas/03_views.sql
```diff
@@ -132,3 +132,104 @@ select count(sub.id) as is_initialized
...
+    ('task.' || t.id) as id,
+    'tasks' as resource,
+    t.id as record_id,
+    t.text as title,
+    concat_ws(' ', tco.first_name, tco.last_name) as subtitle,
+    concat_ws(' ', t.text, t.type) as content,
+    tco.company_id,
+    t.contact_id,
+    null::bigint as deal_id,
+    t.due_date as date
```
pysnooLab: **Future dates poison the shared sort key.** `date` is the shared sort key and every other arm feeds a past timestamp, while tasks feed a future `due_date`. With `date DESC` and a hard `perPage: 50`, tasks crowd out everything: "sourdough" matches 6221 rows across 3 resources, the 50 returned are 50 tasks dated 2027.

Use `coalesce(t.done_date, t.created_at)`.

supabase/schemas/03_views.sql
```diff
@@ -132,3 +132,104 @@ select count(sub.id) as is_initialized
...
+    left join public.companies dc on dc.id = d.company_id
+where d.archived_at is null
+union all
+select
+    ('task.' || t.id) as id,
+    'tasks' as resource,
+    t.id as record_id,
+    t.text as title,
+    concat_ws(' ', tco.first_name, tco.last_name) as subtitle,
+    concat_ws(' ', t.text, t.type) as content,
```
pysnooLab: **Completed tasks are indexed.** No `where` here, unlike the deals arm at L185: 15000 rows with `done_date is not null` are indexed. With the `due_date` sort key above, finished work outranks live records.

`where t.done_date is null`.

supabase/schemas/03_views.sql
```diff
@@ -132,3 +132,104 @@ select count(sub.id) as is_initialized
...
+        ' ',
+        dn.text,
+        (select string_agg(a ->> 'title', ' ') from unnest(dn.attachments) as a)
+    ) as content,
+    nd.company_id,
+    null::bigint as contact_id,
+    dn.deal_id,
+    dn.date
+from public.deal_notes dn
+    left join public.deals nd on nd.id = dn.deal_id;
```
pysnooLab: **Notes on archived deals are indexed.** The deals arm filters `archived_at`, this one does not, and the result links to a deal hidden from the pipeline.

`where nd.archived_at is null`.

supabase/schemas/03_views.sql
```diff
@@ -132,3 +132,104 @@ select count(sub.id) as is_initialized
...
 ) sub;
+
+create or replace view public.search_index with (security_invoker = on) as
+select
+    ('company.' || c.id) as id,
+    'companies' as resource,
+    c.id as record_id,
+    c.name as title,
+    c.sector as subtitle,
+    concat_ws(' ', c.name, c.description, c.website, c.city) as content,
```
pysnooLab: **Indexed fields diverge from the list pages, in both directions.** Companies lack `phone_number`/`zipcode`/`state_abbr` and gain `description`, contacts lack `background`, deals lack `category` and gain the company name. A record findable from its list page is not findable globally.

Derive both from one shared field list per resource.

supabase/schemas/03_views.sql
```diff
@@ -132,3 +132,104 @@ select count(sub.id) as is_initialized
 from (
     select sales.id from public.sales limit 1
 ) sub;
+
+create or replace view public.search_index with (security_invoker = on) as
```
pysnooLab, line 136: **Unindexable by construction.** `ilike '%term%'` over a computed `concat_ws` in a 6-way `UNION ALL` cannot use an index, and `pg_trgm` is not installed. Measured on 95k rows: 6 Seq Scans, 155 ms, plus 98 ms because `count=exact` re-scans for the total. Per keystroke.

Fine at demo scale. Add generated `tsvector` columns plus GIN indexes, or note the limit in the PR.

src/components/atomic-crm/search/GlobalSearchDialog.tsx
```diff
@@ -0,0 +1,120 @@
...
+  onOpenChange,
+}: {
+  open: boolean;
+  onOpenChange: (open: boolean) => void;
+}) => {
+  const translate = useTranslate();
+  const navigate = useNavigate();
+  const isMobile = useIsMobile();
+  const [query, setQuery] = useState("");
+  const { results, isPending, debouncedQuery } = useGlobalSearch(query);
```
pysnooLab: **Failures render as "no result".** `error` is returned by the hook and not destructured, so a missing view or a network failure lands in `CommandEmpty`. No fallback exists: `useGetList` sets `onError = noop` and `checkError` only rejects on 401/403. The user concludes the data does not exist.

Render an error state.

src/components/atomic-crm/search/GlobalSearchDialog.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  // Results whose resource has no page on this layout are dropped rather than
+  // rendered as dead entries.
+  const linkedResults = useMemo(
```
pysnooLab, line 44: **Mobile drops rows after the server-side cap.** The 50-row cap is server-side, then `deals`/`deal_notes` are dropped here because `MobileAdmin` registers only `contacts`, `companies`, `tasks`. When deals dominate the 50 most recent matches, mobile shows zero for a query that works on desktop.

Filter server-side with `resource@in`.

src/components/atomic-crm/search/GlobalSearchButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+  variant?: "labelled" | "icon";
+}) => {
+  const translate = useTranslate();
+  const [open, setOpen] = useState(false);
+
+  useEffect(() => {
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (event.key === "k" && (event.metaKey || event.ctrlKey)) {
+        event.preventDefault();
+        setOpen((previousOpen) => !previousOpen);
```
pysnooLab: **Cmd+K close leaves the query in place.** The shortcut toggles this component's state, bypassing the dialog's `handleOpenChange`, the only caller of `setQuery("")` (Escape and overlay click do clear). The stale term survives, and on reopen it is already past `MIN_SEARCH_LENGTH` so the old results render immediately.

Open only, or route through `handleOpenChange`.

src/components/atomic-crm/search/GlobalSearchButton.tsx
```diff
@@ -0,0 +1,52 @@
...
+  variant = "labelled",
+}: {
+  variant?: "labelled" | "icon";
+}) => {
+  const translate = useTranslate();
+  const [open, setOpen] = useState(false);
+
+  useEffect(() => {
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (event.key === "k" && (event.metaKey || event.ctrlKey)) {
```
pysnooLab: **The shortcut is captured app-wide, including inside inputs.** Bound on `document` with an unconditional `preventDefault()` and no target check: Ctrl+K in a note editor opens the dialog.

Bail out on input, textarea and contenteditable.

src/components/atomic-crm/search/useGlobalSearch.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Query the `search_index` view for a debounced search term.
+ *
+ * Stays idle until the term reaches MIN_SEARCH_LENGTH, so opening the dialog
+ * does not fire a request.
+ */
+export const useGlobalSearch = (query: string) => {
+  const debouncedQuery = useDebouncedValue(query.trim(), DEBOUNCE_MS);
+  const isEnabled = debouncedQuery.length >= MIN_SEARCH_LENGTH;
+
+  const { data, isPending, error } = useGetList<SearchResult>(
```
pysnooLab, line 45: **`content` is fetched on every keystroke and never used.** No `meta.columns`, so `select=*` ships the full `content` blob on 50 rows per keystroke: serialized `email_jsonb`/`phone_jsonb` for contacts, the whole note body for notes (duplicating `title`, itself untruncated server-side). Nothing reads it, `shouldFilter={false}` included.

Set `meta: { columns: [...] }`.

src/components/atomic-crm/dashboard/MobileDashboard.tsx
```diff
@@ -27,6 +28,7 @@ const Wrapper = ({ children }: { children: React.ReactNode }) => {
           />
           <h1 className="text-xl font-semibold">{title}</h1>
         </div>
+        <GlobalSearchButton variant="icon" />
```
pysnooLab: **Search is unreachable from every other mobile screen.** Only mobile call site. `MobileHeader` has 8 consumers, `MobileLayout` has no header, and Cmd+K does not exist on a phone.

Move the trigger into `MobileHeader`.

src/components/atomic-crm/providers/commons/search.ts
```diff
@@ -0,0 +1,176 @@
...
+      record_id: contact.id,
+      title: contactName(contact),
+      subtitle: contact.company_name ?? null,
+      content: null,
+      company_id: contact.company_id ?? null,
+      contact_id: contact.id,
+      deal_id: null,
+      date: contact.first_seen ?? null,
+    })),
+    ...deals
```
pysnooLab: **Archived deals consume slots before being filtered.** The fetch at L22 is unfiltered, so archived deals take places in the top 50 and are only dropped here. The view filters inside the UNION arm, before the limit. Live older deals are lost in the demo build.

Add `"archived_at@is": null` to the deals filter.

src/components/atomic-crm/providers/commons/search.ts
```diff
@@ -0,0 +1,176 @@
...
+      .filter((deal) => !deal.archived_at)
+      .map((deal) => ({
+        id: `deal.${deal.id}`,
+        resource: "deals" as const,
+        record_id: deal.id,
+        title: deal.name,
+        subtitle: null,
```
pysnooLab: **Deals lose their subtitle in the demo build.** The view selects `dc.name as subtitle`; hardcoded `null` here, so the row renders differently per provider. `getLabels` already does this for tasks and notes.

src/components/atomic-crm/providers/commons/search.ts
```diff
@@ -0,0 +1,176 @@
...
+      id: `contact.${contact.id}`,
+      resource: "contacts" as const,
+      record_id: contact.id,
+      title: contactName(contact),
+      subtitle: contact.company_name ?? null,
+      content: null,
+      company_id: contact.company_id ?? null,
+      contact_id: contact.id,
+      deal_id: null,
+      date: contact.first_seen ?? null,
```
pysnooLab, line 135: **Contacts are selected by one key and sorted by another.** L73 fetches the top 50 contacts by `last_seen` but stamps `first_seen`, which is what `byDateDesc` sorts on. The view uses `first_seen` throughout.

src/components/atomic-crm/providers/commons/search.ts
```diff
@@ -0,0 +1,176 @@
...
+      subtitle: dealLabels.get(note.deal_id) ?? null,
+      content: null,
+      company_id: null,
+      contact_id: null,
+      deal_id: note.deal_id,
+      date: note.date ?? null,
+    })),
+  ];
+
+  return rows.sort(byDateDesc).slice(0, MAX_RESULTS);
```
pysnooLab: **The two providers return different result sets, not just different fields.** This path takes 50 per resource and caps at 250; Supabase takes one global top 50. The demo build can surface a record production never returns. Worth adding to the FIXME at L57.

src/components/atomic-crm/providers/supabase/dataProvider.ts
```diff
@@ -353,6 +353,25 @@ const lifeCycleCallbacks: ResourceCallbacks[] = [
...
     },
   },
+  {
+    resource: "search_index",
+    beforeGetList: async (params) => {
+      if (!params.filter?.q) {
+        return params;
+      }
+      const { q, ...filter } = params.filter;
+      // A single `ilike` column: ra-data-postgrest splits the term on spaces
```
pysnooLab, line 366: **The comment describes output the provider does not produce.** `qs.stringify` uses indices format, so two words give `content[0]=…&content[1]=…` and three nest further. It works because PostgREST strips `[n]` subscripts and ANDs repeats, not because repeated `content=` params are sent.

Reword, the next reader will rely on it.

pysnooLab (commented):
Nice shape overall: reusing the `activity_log` `UNION ALL` pattern with `security_invoker = on` is the right call, the migration matches `03_views.sql` exactly so there is no phantom diff, and the `CommandDialog` fix (moving `DialogHeader` inside `DialogContent`) is a genuine a11y improvement. Findings below were verified against a local Postgres and a running PostgREST rather than read off the diff.

Three of them break the feature's own stated use case, so worth a look before merge:

1. **Multi-word search fails on 4 of the 6 arms.** The parent name is displayed as `subtitle` but is absent from `content`, and the provider ANDs one `ilike` per word. "sourdough Thomas" returns nothing for a note on Thomas Anderson containing "sourdough".
2. **Tasks monopolise the results.** `due_date` is a future timestamp feeding the shared `date DESC` sort key, against past timestamps everywhere else, under a hard 50-row cap with no pagination. Measured: 6221 matches across 3 resources, 50 tasks returned, 0 of anything else. The tasks arm also has no `where`, so completed tasks are indexed too.
3. **Errors are indistinguishable from empty results.** The `error` from `useGlobalSearch` is unused, so a missing view or a network failure renders "No result found" with no notification.

**Missing e2e coverage.** `e2e/` is untouched. The two added tests are welcome but cover only the FakeRest path: not the Supabase `content@ilike` filter, not Cmd+K, not navigation on select. The repo's `e2e-conventions` rule asks for an e2e spec on changes touching UI, filters, search or keyboard interactions, and exempts only pure CSS and migration-only changes. One spec asserting "a note word plus the parent contact name returns the note" would have caught findings 1, 2 and the mobile one at once.

Also worth noting: the docs checkbox is unchecked, and the behavior deserves a line in the user docs (min 2 characters, 50-row cap, no pagination, which resources are covered).

pysnooLab (changes requested):

supabase/schemas/03_views.sql
erwanMarmelab: `tasks` has no `created_at` (01_tables.sql:111). Added the column instead,
backfilled with `coalesce(done_date, due_date)` — same shape as `companies`
and `deals`. Verified after a full reset: a task due 2027-09-10 indexes at
2026-08-11.

supabase/schemas/03_views.sql
erwanMarmelab: Taking your second option: noting the limit, not fixing it. tsvector + GIN
needs a migration on all six tables and breaks FakeRest parity — follow-up if
volume grows. The 50-row cap is in the user docs.

supabase/schemas/03_views.sql
erwanMarmelab: Aligned both ways: +phone_number/zipcode/state_abbr, +background, +category.
Not derived from a shared list though — the list search is a TS array, the
view is SQL. Left a comment pointing at the TS side.

supabase/migrations/20260910140459_global_search.sql
```diff
@@ -0,0 +1,83 @@
+alter table "public"."tasks" add column "created_at" timestamp with time zone not null default now();
+
+update "public"."tasks" set "created_at" = coalesce("done_date", "due_date", "created_at");
```
pysnooLab, line 3: The view filters where t.done_date is null, so for every task row it exposes the done_date branch is unreachable and created_at always resolves to due_date. Open tasks then carry future dates, and since search_index is sorted date DESC with a hard perPage: 50 and no pagination or ranking, they outrank every other result type. Grouping in the dialog is client-side, so it cannot recover rows the server never returned.

Not visible locally: a fresh install has an empty tasks table and the FakeRest generator uses past dates. It only shows up on existing deployments after upgrade.

Suggest coalesce("done_date", "created_at"), or dropping the backfill.

src/components/atomic-crm/providers/commons/search.ts
```diff
@@ -0,0 +1,185 @@
...
+  return data as T[];
+};
+
+const getLabels = async <T extends { id: Identifier }>(
+  dataProvider: DataProvider,
+  resource: string,
+  ids: (Identifier | undefined | null)[],
+  getLabel: (record: T) => string,
+): Promise<Map<Identifier, string>> => {
+  const uniqueIds = [...new Set(ids.filter(Boolean) as Identifier[])];
```
pysnooLab, line 39: nit : ids.filter(Boolean) drops the valid id 0. FakeRest generators assign array-index ids (Array.from(Array(size).keys())) and FakeRest preserves them (addOne checks identifier != null), so record 0 exists in the demo build. Consequence is a missing subtitle on tasks and notes belonging to contact 0 or deal 0. Suggest ids.filter((id) => id != null).

Same pattern at line 142 for deal.company_id. Demo-only: Supabase identity columns start at 1.

pysnooLab (changes requested):

supabase/migrations/20260910140459_global_search.sql
erwanMarmelab: Right, the done_date branch was dead — thanks.

Not taking coalesce(done_date, created_at) though: it lands every open task
on the migration instant, so they all tie at the top of the same date DESC
sort. Same crowding, dated today instead of 2027.

Now trusting due_date only once it has passed, falling back to the contact's
first_seen, clamped with least(..., now()). Measured on an open task due
2027-09-10: indexes at 2026-02-27 instead of the migration day.

fzaninotto:
please let me test this one before merge

PR #358, opened 2026-09-21
Feat: update to shadcn-admin-kit v2

erwanMarmelab:
Upgrade to shadcn admin kit v2

## Additional Checks


vitest.config.ts
```diff
@@ -16,7 +17,7 @@ export default defineConfig({
   test: {
     projects: [
       {
-        plugins: [react()],
+        plugins: [react(), tailwindcss()],
```
erwanMarmelab, line 20: The stylesheet is required, not cosmetic: Base UI's modal backdrop is inline-styled `position:fixed;inset:0`, while the dialog popup gets its stacking from Tailwind classes alone. Without CSS the backdrop covers the popup and every click inside a dialog or sheet is swallowed by `<div data-base-ui-inert>`

src/components/atomic-crm/layout/MobileNavigation.tsx
```diff
@@ -94,17 +94,16 @@ const NavigationButton = ({
...
 }) => (
   <Button
-    asChild
     variant="ghost"
     className={cn(
       "flex-col gap-1 h-auto py-2 px-1 rounded-md w-16",
       isActive ? null : "text-muted-foreground",
     )}
+    render={<Link to={href} />}
+    nativeButton={false}
```
pysnooLab: `nativeButton={false}` makes Base UI add `role="button"`, so this link is announced as a button (it was a plain link with Radix `asChild`). Same in `DealEdit`, `CompanyShow`, `HotContacts`, `authentication.tsx` and `ui/pagination.tsx`. Suggest `<Link className={buttonVariants(...)}>` instead, like the "Add contact" link in `DashboardStepper`.

src/components/atomic-crm/dashboard/DashboardStepper.tsx
```diff
@@ -139,18 +138,23 @@ export const DashboardStepper = ({
...
+                      render={
+                        <Link
+                          to={createPath({
+                            resource: "contacts",
+                            type: "show",
+                            id: contactId,
+                          })}
+                        />
+                      }
+                      nativeButton={false}
```
pysnooLab: Exception: this one needs `disabled`, which Base UI handles on a link, so `role="button"` is fine here.

src/components/admin/autocomplete-array-input.tsx
```diff
@@ -144,122 +162,163 @@ export const AutocompleteArrayInput = (
     [inputText, getChoiceText],
   );
 
+  const handleChange = useEvent((choice: any) => {
+    setFilterValue("");
+    if (isFromReference) {
+      setFilters(filterToQuery(""));
+    }
+    const value = getChoiceValue(choice);
```
pysnooLab: When `onCreate` returns nothing, ra-core passes the create item here, and `"@@ra-create"` gets saved in the value (no badge is shown). Suggested guard:

```suggestion
    const value = getChoiceValue(choice);
    if (value === (createValue ?? "@@ra-create")) {
      return;
    }
```

pysnooLab (changes requested):

